PR #9, opened 2015-06-23
Use correct windres when building flexlink.exe

dra27:
Small correction to the Makefile so that flexlink.exe can be built on machines lacking default gcc


alainfrisch:
Thanks David!

PR #10, opened 2015-06-23
Include version number in source tarball directory

dra27:
Is there a reason why you use `flexdll` rather than `flexdll-$(VERSION)` in the `package_src` target in the Makefile, given that the tarball is named flexdll-$(VERSION).tar.gz?
Of the large number of tarballs I extract for my OCaml installation, flexdll is the only one which doesn't include its version in the resulting folder (i.e. matching the tarball)
I _personally_ prefer the version number convention as I sometimes end up wanting multiple versions of a library in the same directory (for whatever reason)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -162,13 +162,13 @@ URL = frisch@frisch.fr:www/flexdll/
 PACKAGE = flexdll-$(VERSION).tar.gz
 
 package_src:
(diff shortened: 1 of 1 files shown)


alainfrisch:
No particular reason, and no problem to do as you suggest.  Thanks!

PR #11, opened 2015-06-25
Do not read aux section when COMDAT selection number is not 5

Nevor:
The aux parameter 12 is unused when the parameter 14 is not 5 according to the [documentation section 5.5.5, page 52](http://www.skyfree.org/linux/references/coff.pdf). This will allow linking of coff files generated by clang-cl.


alainfrisch:
Thanks!

PR #14, opened 2016-01-01
Visual Studio updates

dra27:
Happy New Year! Several patches arising from recent work testing OCaml trunk on various old Visual Studio installations:

 - Previously added `.gitattributes` omitted the important normalisation of existing files (sorry!)
 - It is still possible to build OCaml on some old but just about supported Microsoft compilers whose development was before Microsoft's major security overhaul for Windows XP SP2 and so which don't include the secure versions of CRT functions (this includes `msvcrt.dll` used by the mingw ports). Corrected the `#ifdef` logic to share the mingw branch with older Microsoft C compilers too. This minor adjustment enables compilation in Microsoft Visual Studio .NET 2003 and also in the Windows Server 2003 SP1 SDK (which is a pre-release Visual Studio 2005 compiler).
 - Two changes which aid bootstrapping FlexDLL from within OCaml (this essentially extends support already required for building Cygwin OCaml from sources). `MSVC_DETECT` allows all of the complex logic for detecting both 32 and 64-bit Microsoft C compilers to be disabled in favour of just trusting the environment (as for an OCaml compilation). The bootstrapping process shortly to become an OCaml PR allows a bytecode flexlink.exe to be compiled before starting `make -f Makefile.nt world`, hence `.cmo` artefacts may now be seen. The `checkenv` script (which must be run manually) informs the user if it looks safe to run `make MSVC_DETECT=0 CHAINS=msvc[64] ...`


alainfrisch:
Thanks!

PR #15, opened 2016-01-18
Prevent MSVC displaying .c filename when compiling

dra27:
See also https://github.com/ocaml/ocaml/pull/407.

### DIFF
--- a/reloc.ml
+++ b/reloc.ml
@@ -1107,16 +1107,39 @@ let setup_toolchain () =
   | `LIGHTLD ->
       search_path := !dirs
 
+let display_msvc_output file name =
+  let c = open_in file in
+  try
+    let first = input_line c in
+    if first <> name then
+      print_string first;
+    while true do
+      print_string (input_line c)
+    done
+  with _ ->
+    close_in c
+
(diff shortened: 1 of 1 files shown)


alainfrisch:
Thanks!

dra27:
Ah - we're working too quickly - something's failing in the testsuite run... give me a minute!

dra27:
(sorry, I thought enough of the testsuite had run...)

alainfrisch:
Is that a problem with the PR on flexdll, or the one in OCaml?

dra27:
This one on FlexDLL - I accidentally shadowed a variable, but it doesn't pick up when building OCaml because it never exercise ocamlc -custom

dra27:
I'm just (properly) testing the actual fix - do you want me to push an additional PR, or will you reset to HEAD~1 on master (never thought I'd write that!)

alainfrisch:
I prefer an additional PR.

dra27:
OK - immortalise my embarrassing errors, why don't you :wink: It's done - but I'm allowing a full testsuite run before opening the PR this time!

PR #16, opened 2016-01-18
Fix errors in previous attempt to silence MSVC

dra27:
Correct fix this time!

### DIFF
--- a/reloc.ml
+++ b/reloc.ml
@@ -1111,7 +1111,7 @@ let display_msvc_output file name =
   let c = open_in file in
   try
     let first = input_line c in
-    if first <> name then
+    if first <> (Filename.basename name) then
       print_string first;
     while true do
       print_string (input_line c)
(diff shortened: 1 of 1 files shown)


alainfrisch:
Is it intentional to not show the content of the redirected stdout in case of error?

dra27:
No!

dra27:
That should deal with that (I'm just about to run it) - but one other question. I don't think it's my regression, but in dry_mode, should it be compiling the file??

alainfrisch:
> but in dry_mode, should it be compiling the file??

No, of course!

dra27:
I wasn't sure, given how that it works - in which case, I'll fix that while I'm here!

dra27:
There we go - that should fix that too. x86 and x64 builds and testsuite presently running...

dra27:
That all seems to be fine

alainfrisch:
Thanks!

PR #17, opened 2016-04-06
Improve build system to detect available compilers

dra27:
Refactor some configuration items from `Makefile`


Makefile
```diff
@@ -1,41 +1,14 @@
...
-	echo "let version = \"$(VERSION)\"" > version.ml
-	echo "let mingw_prefix = \"$(MINGW_PREFIX)\"" >> version.ml
-	echo "let mingw64_prefix = \"$(MINGW64_PREFIX)\"" >> version.ml
-
-# Supported tool-chains
-
-CHAINS = mingw mingw64 cygwin cygwin64 msvc msvc64
-
-# Compilers
+ifeq ($(findstring clean,$(MAKECMDGOALS)),)
```
dra27: Avoids running `configure` if `make [dist]clean` is called twice.

Makefile
```diff
@@ -1,41 +1,14 @@
...
+include Makefile.config
+endif
 
-# Attempt to locate the Windows SDK
+all: $(if $(filter-out no,$(HAVE_OCAMLOPT)),flexlink.exe) support
```
dra27: This has the nice benefit if bootstrapping just the FlexDLL objects that `make all` only builds what it can.

Makefile
```diff
@@ -1,41 +1,14 @@
...
 
-ifeq ($(findstring clean,$(MAKECMDGOALS)),)
-include Makefile.winsdk
+ifneq ($(filter-out no,$(HAVE_OCAMLOPT)),)
```
dra27: Eliminates the warning about `Makefile.config` not being found (and ocamlopt not being found) when bootstrapping.

Makefile
```diff
@@ -190,10 +143,11 @@ demo_msvc64:  flexlink.exe flexdll_msvc64.obj flexdll_initer_msvc64.obj
 	(cd test && $(MSVC64_PREFIX) $(MAKE) clean demo CHAIN=msvc64 CC="$(MSVCC64)" O=obj)
 
 distclean: clean
-	rm -f Makefile.winsdk
+	rm -f Makefile.config config.log config.status version.ml
+	rm -rf autom4te.cache
 
 clean:
-	rm -f *.obj *.o *.lib *.a *.exe *.opt *.cmx *.dll *.exp *.cmi *.cmo *~ version.res version.ml
+	rm -f *.obj *.o *.lib *.a *.exe *.opt *.cmx *.dll *.exp *.cmi *.cmo *~ version.res
```
dra27: `version.ml` is generated by `configure`, so should be removed in `distclean` only.

configure.ac
```diff
@@ -0,0 +1,57 @@
+AC_INIT(FlexDLL,0.35,alain@frisch.fr,flexdll,http://alain.frisch.fr/flexdll.html)
+#AC_COPYRIGHT(Copyright 2016 ?)
+
+AC_ARG_VAR([MINGW_PREFIX],[Prefix for the mingw-w64 i686 cross-compiler])
+AC_ARG_VAR([MINGW64_PREFIX],[Prefix for the mingw-w64 x64 cross-compiler])
+AS_IF([test x"${MINGW_PREFIX}" = x""],[MINGW_PREFIX=i686-w64-mingw32])
+AS_IF([test x"${MINGW64_PREFIX}" = x""],[MINGW64_PREFIX=x86_64-w64-mingw32])
```
dra27: I have used `AC_ARG_VAR` as it allows the prefix for the mingw compilers to be overridden. I've done that only because they are written to `version.ml`, but I'm not sure why/when you'd want to override this anymore (was it support the older mingw compilers too?).

configure.ac
```diff
@@ -0,0 +1,57 @@
+AC_INIT(FlexDLL,0.35,alain@frisch.fr,flexdll,http://alain.frisch.fr/flexdll.html)
```
dra27: Note that this is where the version number is now declared!

configure.ac
```diff
@@ -0,0 +1,57 @@
...
+
+# Check for Microsoft compilers
+AC_MSG_CHECKING([for an installed pair of Microsoft C Compilers])
+MSVS_PATH=no
+eval `./msvs-detect --with-assembler --with-mt`
+MS_CHAINS=
+AS_IF([test "${MSVS_PATH}" = "no"],[
+  AC_MSG_RESULT([no])
+  AC_MSG_CHECKING([for an installed x86 Microsoft C Compiler])
+  eval `./msvs-detect --with-assembler --with-mt --arch=x86`
```
dra27: It is extremely difficult to install an x64 compiler only, so there's no test for that weird case!

PR #18, opened 2016-04-28
Moved add file logic to tr functions

bschommer:
The ocaml compiler also writes options in the @response which lead
to problems when flexlink tries to find the files. See mantis GPR#6864


alainfrisch:
Thanks!

PR #20, opened 2016-07-17
Remove implicit field name in record constructions

dra27:
Don't use the let field = () in {field} feature (3.12.0 onwards)

### DIFF
--- a/coff.ml
+++ b/coff.ml
@@ -336,7 +336,7 @@ module Symbol = struct
       section = `Num (int16 buf 12);
       stype = int16 buf 14;
       storage = int8 buf 16;
-      auxn;
+      auxn = auxn;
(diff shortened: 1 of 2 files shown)


dra27:
I'd like to maintain packages where flexlink is compiled in older OCamls. The most invasive change is not using OCaml 3.12.0's implicit record field names feature. This one would be useful, as it affects far more lines of code, increasing the likelihood of the patch needing regenerating over time...
The rest requires not using `{|...|}` in `cmdline.ml` and inserting some Bytes-related compatibility declarations in `coff.ml` but I imagine you're not particularly interested in those two changes?

alainfrisch:
Well, I don't want to commit to maintaining compatibility with older OCamls, but if you feel like doing so, I'll certainly accept such PRs.

PR #21, opened 2016-07-20
Add build compatibility for 3.11.0--4.02.3

dra27:
Apropos of #20, this commit adds complete build compatibility for 3.11.0 onwards. My opinion - which I expect matches yours - is that this can be kept until such point as a language feature is used where back-porting becomes non-trivial, and then it can be completely removed.

The code in `StringCompat.ml` is lifted straight out of OCaml 4.03.0 (I'm assuming licence compatibility/permission to do that?)

`Makefile` tests ocamlopt -version each time and uses a cookie to detect changes in the compiler (for example, if using opam to switch between multiple compilers...) and regenerates the files.

I've tested it with 3.11.0, 4.02.3 and 4.03.0 - it'll get a bit more battle testing shortly!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@ VERSION = 0.35
 all: flexlink.exe support
 
 include $(shell cygpath -ad "$(shell ocamlopt -where)/Makefile.config")
+OCAML_VERSION:=$(shell ocamlopt -version|sed -e "s/+.*//" -e "s/\.//g")
 
 MINGW_PREFIX = i686-w64-mingw32-
 MINCC = $(MINGW_PREFIX)gcc
@@ -15,8 +16,7 @@ CYGCC = $(CYGWIN_PREFIX)gcc
 CYGWIN64_PREFIX = x86_64-pc-cygwin-
 CYG64CC = $(CYGWIN64_PREFIX)gcc
 
-.PHONY: version.ml
-version.ml:
+version.ml: Makefile
 	echo "let version = \"$(VERSION)\"" > version.ml
 	echo "let mingw_prefix = \"$(MINGW_PREFIX)\"" >> version.ml
 	echo "let mingw64_prefix = \"$(MINGW64_PREFIX)\"" >> version.ml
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -118,10 +118,41 @@ build_cygwin64: flexdll_cygwin64.o flexdll_initer_cygwin64.o
...
+
+# Enable BytesCompat module for OCaml < 4.02.0
+ifeq ($(call test_ver,4020),lt)
+BytesCompat.ml: BytesCompat.ml.in COMPILER-$(OCAML_VERSION)
+else
+BytesCompat.ml: COMPILER-$(OCAML_VERSION)
+	touch $@
+endif
+
+cmdline.ml: cmdline.ml.in COMPILER-$(OCAML_VERSION)
```
alainfrisch: Let's just use a regular string in cmdline.ml.

Makefile
```diff
@@ -196,7 +227,7 @@ distclean: clean
 	rm -f Makefile.winsdk
 
 clean:
-	rm -f *.obj *.o *.lib *.a *.exe *.opt *.cmx *.dll *.exp *.cmi *.cmo *~ version.res version.ml
+	rm -f *.obj *.o *.lib *.a *.exe *.opt *.cmx *.dll *.exp *.cmi *.cmo *~ version.res version.ml COMPILER-* cmdliner.ml BytesCompat.ml StringCompat.ml
```
alainfrisch: Typo cmdliner.ml (extra 'r'), although this would go away according to the previous note.

Makefile
dra27: New commit!

Makefile
dra27: Oops - though, indeed, it's gone now!

dra27: I long for properly working eyesight again! 👓

alainfrisch:
The compatibility modules should be renamed to something like Compat402, Compact403 since they are not about one specific module in stdlib but rather about adding a compatibility layer for a specific old version.

dra27:
OK - technically speaking the Compat402 (what is presently BytesCompat) contains a definition of `|>` which is actually Compat401 - I assume you're OK with that, rather than adding a third compatibility module for that one-liner?

alainfrisch:
> I assume you're OK

Yes.

I was also wondering if it would simplify the Makefile to simply concatenate in a rule the required Compat* (depending on the version of OCaml) into a single Compat.ml, and include that one in coff.ml.

dra27:
Yes, it does simplify the Makefile greatly!

Makefile
```diff
@@ -238,7 +222,7 @@ package_src:
 	rm -Rf flexdll-$(VERSION)
 	mkdir flexdll-$(VERSION)
 	mkdir flexdll-$(VERSION)/test
-	cp -a *.ml Makefile $(COMMON_FILES) version.rc flexdll-$(VERSION)/
+	cp -a $(filter-out Compat.ml,$(wildcard *.ml)) Makefile $(COMMON_FILES) version.rc flexdll-$(VERSION)/
```
dra27: I noticed when preparing the tags the other day that the inclusion of version.ml in the sources tarball is not explicit - it gets picked up if there's a build already there). Perhaps `package_src` should depend on `version.ml`? The wildcard has also occasionally included some debugging files (`bla.ml` and `bug.ml` :smile:)

dra27:
It would be cleaner I think to squash these three commits before merging? Let me know if you're happy with the changes, and I'll force push a single squashed commit to this PR.

alainfrisch:
Yes, the current set looks very good.  Can you indeed squash and add a line in the changelog?  Thanks!

Makefile
alainfrisch: Do we actually need to include version.ml in source tarball?  It should be generated during the build.

And yes, an explicit of files to include would be much better of course...

dra27:
I edited your line in `CHANGES` for the modernisation, rather than adding a separate one.

dra27:
Further fixes to the `package_src` target probably clearer in a separate (tiny!) PR

alainfrisch:
Thanks again!

PR #22, opened 2016-07-20
Explicitly state .ml files for sources tarball

dra27:
Prevent any working .ml files being included in the sources
distribution.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -222,7 +222,7 @@ package_src:
 	rm -Rf flexdll-$(VERSION)
 	mkdir flexdll-$(VERSION)
 	mkdir flexdll-$(VERSION)/test
-	cp -a $(filter-out Compat.ml,$(wildcard *.ml)) Makefile $(COMMON_FILES) version.rc flexdll-$(VERSION)/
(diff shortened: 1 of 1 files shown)


dra27:
One line fix :smile:

Makefile
```diff
@@ -222,7 +222,7 @@ package_src:
 	rm -Rf flexdll-$(VERSION)
 	mkdir flexdll-$(VERSION)
 	mkdir flexdll-$(VERSION)/test
-	cp -a $(filter-out Compat.ml,$(wildcard *.ml)) Makefile $(COMMON_FILES) version.rc flexdll-$(VERSION)/
+	cp -a $(filter-out Compat.ml version.ml,$(OBJS) $(shell git ls-files Compat*.ml)) Makefile $(COMMON_FILES) version.rc flexdll-$(VERSION)/
```
dra27: Use of `git ls-files` instead of `$(wildcard ...)` has the desirable side-effect that additional Compat files which haven't been added to git are *not* included in the sources tarball.

alainfrisch:
Thanks!

PR #23, opened 2016-08-02
Remove warnings from format strings.

bschommer:
MSVC emits some warnings for wrong format strings. Adding a few
appropriate casts avoids them.


alainfrisch:
Thanks!

PR #24, opened 2016-08-25
Ensure bootstrapping picks up correct version

dra27:
Calling ocamlopt directly could pick up a system-installed ocamlopt, and
so select the wrong version of the compiler.


dra27:
I wouldn't update the submodule in OCaml for this - all that can happen is that the compatibility code might be unnecessarily used. It's better for the submodule in OCaml to represent a tagged release of FlexDLL.

alainfrisch:
Well, spotted!

dra27:
Sadly it was my line of code in the first place!

PR #25, opened 2016-09-12
Further fix to bootstrapping support

dra27:
Problems caused further down `Makefile` if `$(OCAMLOPT)` is empty!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,11 @@ VERSION = 0.35
 all: flexlink.exe support
 
 include $(shell cygpath -ad "$(shell ocamlopt -where)/Makefile.config")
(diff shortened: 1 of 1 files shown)


dra27:
Actually, #24 broke compiling flexlink in the normal way, and this doesn't fix that. I'll revert that commit, fix it properly and then push that here hopefully tomorrow (i.e. don't merge this PR as it stands)

dra27:
Right - that seems to fix it properly! `OCAMLOPT` supplied with a default so that it builds normally (i.e. not bootstrapped) and `OCAML_VERSION` will always have a number, so the error messages should be clearer.

alainfrisch:
Thanks!

PR #26, opened 2016-09-14
Fix sources tarball

dra27:
There is a mistake in `Makefile` from when my findwinsdk script was added - it needs to be included in the sources tarball.

This patch fixes Makefile for future versions - please could you regenerate the .tar.gz file on your website for 0.35 (or just add the findwinsdk script to it)?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -222,7 +222,7 @@ package_src:
 	rm -Rf flexdll-$(VERSION)
 	mkdir flexdll-$(VERSION)
 	mkdir flexdll-$(VERSION)/test
-	cp -a $(filter-out Compat.ml version.ml,$(OBJS) $(shell git ls-files Compat*.ml)) Makefile $(COMMON_FILES) version.rc flexdll-$(VERSION)/
+	cp -a $(filter-out Compat.ml version.ml,$(OBJS) $(shell git ls-files Compat*.ml)) Makefile findwinsdk $(COMMON_FILES) version.rc flexdll-$(VERSION)/
 	cp -aR test/Makefile test/*.c flexdll-$(VERSION)/test/
 	tar czf $(PACKAGE) flexdll-$(VERSION)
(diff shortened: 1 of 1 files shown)


alainfrisch:
Did you find any public link to that .tar.gz?  The README now points to source releases created by GitHub.

dra27:
I guessed it - the OPAM packages I've put together use it. Are they deprecated?

alainfrisch:
I haven't given it much thought, but GitHub tarballs are rather convenient, and they are the ones currently advertised in the README; I'm not sure I will continue to upload .tar.gz to my server.

dra27:
In which case, would it be best to remove the tarball lines completely Makefile?

alainfrisch:
I'm ra

dra27:
In postponing, please could the findwinsdk file be added to the tarball too - then I can also postpone updating the opam files :wink:

alainfrisch:
I've updated the uploaded tarball, although I'm not sure you can get away without updating the opam file (which contains a md5sum of the tar.gz, no?).

dra27:
Thanks - that means I just update the 0.35 opam file... if the switch to GitHub tarballs becomes official, there're the 0.29-0.34 ones to update too!

dra27:
I just tried it - unfortunately you've added the file to the root, rather than in `flexdll-0.35/`. This'll fix it:
```
gunzip flexdll-0.35.tar.gz
tar -xf flexdll-0.35.tar findwinsdk
tar -f flexdll-0.35.tar --delete findwinsdk
tar -rf flexdll-0.35.tar --transform="s|^|flexdll-0.35/|" --remove-files findwinsdk
gzip flexdll-0.35.tar
```

alainfrisch:
Oops, fixed.

dra27:
Marvellous, ta!

Issue #7, opened 2015-06-02
Command line too long


gares:
I'm using cygwin, the mingw64 toolchain and I get the error.
It comes from Reloc.cygpath, that in turns calls get_output with use_bash:false.
The quick hack was to force it to use bash, but what could be done (in the non use_bash case) is to use Unix.create_process instead of Sys.command, since the former does not use a shell (no cmd.exe, so more than 8k for argv).  See for example sys_command in https://github.com/coq/coq/blob/trunk/lib/cUnix.ml 

Best,

alainfrisch:
It was initially planned that flexdll should not depend on Unix, to make it easier to integrate it directly in OCaml's bootstrap cycle, but since this is not going to happen, there is no reason not to use it.

I won't have much time to work on that soon, so if you feel like suggesting a patch, I'll happily include it.

gasche:
@gares , I opened an issue in the OCaml bugtracker [#7050](http://caml.inria.fr/mantis/view.php?id=7050) to be able to pass command-line arguments for the OCaml compiler in a file, to avoid argv size limitations. The two issues seem independent, but are in fact related if the command that is called in the present PR is an invocation of the OCaml compiler (or a tool distributed in the OCaml compiler distribution). Is it?

dra27:
Update for 2016: FlexDLL is now an optional part of OCaml's bootstrap cycle, so a hard dependency on Unix would be upsetting (for at least two developers...!). However, it sounds as though this could be done in a way which could be easily disabled at build time in the bootstrap (similarly to the shims already in place for building with OCaml < 4.03.0)

bschommer:
I also will submit a PR in the next few days which allows flexdll to parse responsefiles produced by the new ocaml functions `write_arg` and `write_arg0`. But I'm not that familiar with the bootstrap mechanism to add a check if the flexdll support this..

dra27:
The FlexDLL bootstrap in OCaml is only normally something you have to worry about if you're altering the OCaml side - feel free to cc me when you submit the PR and I'll happily test it. If you want to have a go, it's an easy enough procedure: from a git clone of OCaml (you can do it from a tarball of the sources, but it's easier to use a clone):

 - `git submodule update --init`
 - `cd flexdll`
 - `git remote add -f bschommer https://github.com/bschommer/flexdll.git`
 - `git checkout bschommer/your-new-flexdll-branch [-b your-new-flexdll-branch]`
 - `cd ..`
 - *configure OCaml*
[...]

PR #30, opened 2016-12-20
Added quoting functions for mingw, cygwin, etc.

bschommer:
The new gnu_quote functions is an ocaml reimplementation of the libiberty
writeargv function.
Responsefiles will be create for msvc if the command line lenght is larger
than 1024 and for gnu based tools for 8192.

This should solve issue #7 .

### DIFF
--- a/reloc.ml
+++ b/reloc.ml
@@ -94,6 +94,28 @@ let get_output1 ?use_bash cmd =
 
 let mk_dirs_opt pr = String.concat " " (List.map (fun s -> pr ^ (Filename.quote s)) !dirs)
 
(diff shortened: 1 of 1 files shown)


alainfrisch:
build_diversion is called from quote_files, used to produce the command line for calling the linker through run_command.  When the resulting command line is longer than 8192 characters, this function dumps the command line to a file and applies bash to it.  My understanding was that this is enough to pass arbitrarily long command lines to Cygwin programs (relying on some Cygwin magic).  Is it not the case?

bschommer:
I'm not sure. However it would still be better to call the tool directly since process creation is expensive.

PR #31, opened 2016-12-21
Added -arg,-arg0 functions like in ocaml.

bschommer:
This allows the usage of -arg and -arg0 like for the ocaml compiler.

I will also submit a PR against ocaml such that the ocaml compiler prints the arguments using `write_arg0`.


alainfrisch:
Thanks, merged. (Finally!)

PR #37, opened 2017-07-13
Add AppVeyor script

nojb:
What it does:

1. Compiles OCaml using the `msvc64` toolchain and a precompiled binary of `flexdll 0.34`.
2. Compiles `flexlink.exe` using this OCaml and the support C code with the six supported toolchains.
3. It builds & runs the demo code again with the six supported toolchains.

The OCaml versions tested are `4.02`, `4.03`, and `4.04`.  The binaries are cached so that there is no need to recompile each time.  We need to add `4.05` but need to tweak the build instructions (no more `Makefile.nt`).

Should we be testing `4.01` as well ?

When/if this is merged you will need to connect the repository to AppVeyor to get things started.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,45 @@
+platform:
+  - x64
+
+image: Visual Studio 2015
+
+environment:
+  global:
+    CYG_ROOT: C:/cygwin
+    CYG_MIRROR: http://mirrors.kernel.org/sourceware/cygwin/
+    CYG_CACHE: C:/cygwin/var/cache/setup
+    FLEXDLL_DIR: C:/flexdll
+  matrix:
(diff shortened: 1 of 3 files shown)


nojb:
There is some issue with the CRT version (I think) that I am not sure how to resolve...  Maybe @dra27 has some idea (e.g. https://ci.appveyor.com/project/nojb/flexdll/build/1.0.51#L1801).

nojb:
OK, problem(s) solved - it now tests 4.03, 4.04, 4.05, and trunk (all tests pass).  4.02 needs to be fixed so it is deactivated for now.

alainfrisch:
Wonderful!  Will merge early next week.

Issue #39, opened 2017-08-29
Flexdll is incompatible with visual studio 2017.3


bschommer:
When compiling ocaml with visual studio 2017.3 I get the following error message during linking of the `unix` library:
    
    Creating library C:\cygwin\tmp\dyndll_implib571fa3.lib and object 
     C:\cygwin\tmp\dyndll_implib571fa3.exp
     C:\cygwin\tmp\dyndll_implib571fa3.lib : fatal error LNK1136: invalid or corrupt file

bschommer:
There is a Mantis issue for this + PR so I'm closing this.

PR #40, opened 2017-08-30
Workaround apparent bug in VS 2017.3 link

dra27:
This is very weird - I can't see any in the VS 2017.3 release notes to suggest this change to the linker, but the error reported in #39 can be triggered by empty files, so I tried writing a single byte and the problem disappears!

I believe that the only reason that the implib is specified is so that flexlink can delete the .lib and .exp file cleanly, there are three other solutions:
 1. Sandbox the entire link command in a sub-directory of tmp and erase everything in it at the end (tedious amount of code...)
 2. Create the .exp file instead using `temp_file` and generate the .lib name from that (the linker doesn't seem to care about the .exp file existing)
 3. Create another file entirely and generate the .lib name from that

But I stopped at this solution, as it worked!

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,5 +1,6 @@
 Next Version
 - Add Unicode support (patch by Nicolás Ojeda Bär)
+- Workaround apparent bug in VS 2017.3 link (patch by David Allsopp)
 
 Version 0.35
 - Fixes for clang (patch by Roven Gabriel)
--- a/reloc.ml
+++ b/reloc.ml
@@ -986,6 +986,9 @@ let build_dll link_exe output_file files exts extra_args =
           else
             temp_file "dyndll_implib" ".lib"
         in
+        (* VS 2017.3 doesn't seem to be able to cope with /implib: existing but
+           being an empty file. *)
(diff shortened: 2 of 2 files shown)


alainfrisch:
Great, excellent.  I'm always having cold sweats when I hear about flexdll being broken with new versions of VS...

Do you have an idea why `link` would try to open an existing file matching the implib target?

With the fix, have you been able to compile all of the OCaml distribution successfully with VS 2017.3?

dra27:
It's not just you - this one was more disturbing because it's an update, not a whole new release!

I have tested trunk msvc32 and msvc64, yes.

I can only imagine this has come from some kind of unintended side-effect - I rigged a test, for example, where I copied dllthreads.lib to the temporary file and built dllunix with no problem... the content didn't seem to matter, just that either the file was not present or it had at least 1 byte.

bschommer:
Maybe we should report this to the visual studio team. This seems really kind of strange.

dra27:
Indeed - though we should reduce it to a minimal example first

alainfrisch:
Would by any chance `link /implib:foo.lib ...` fail as soon as foo.lib is an existing empty file?

PR #43, opened 2017-09-27
Tweaks

dra27:
- Cosmetic tidy-up to demo targets
 - Stop on first error in `make support`
 - Check that the correct version of cl is being used (32/64 bit)


alainfrisch:
Thanks, excellent, and very useful!  Waiting for AppVeyor feedback, just in case.

alainfrisch:
AppVeyor is failing on trunk because we compile OCaml using flexdll 0.35, which is not ok.  I guess the good test would be to actually do the OCaml+flexdll bootstrapping together.

PR #47, opened 2017-10-21
Fix encoding of high surrogate in UTF-16

dra27:
Found while torturing AppVeyor by compiling OCaml in a directory called :camel:

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,7 +4,7 @@ Next version
   causing https://caml.inria.fr/mantis/view.php?id=7603
 - Support for passing argument through external files (-arg/-arg0)
   (Bernhard Schommer)
-
+- Fix encoding of high surrogate for U+10000-U+10FFFF in UTF-16 response files (David Allsopp)
 
 Version 0.36
 - Add Unicode support (patch by Nicolás Ojeda Bär)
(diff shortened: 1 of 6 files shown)


dra27:
I'll review #46 tomorrow so that that could be included too, but I think there should be a release pre-4.06.0 containing this fix (it'd be hard to attack, but it feels wrong for OCaml to depend on a program with a faulty Unicode conversion)

nojb:
Unrelated to this PR of course, but the UTF-16 encoding could be done with `Buffer. add_utf_16le_uchar` from 4.06 onwards and we could add the present code to the `Compat` module.

dra27:
Actually, that's a much better idea - I'll add that to tomorrow's train-list!

dra27:
And three more commits to do that:
 - I found on my local AppVeyor an issue with the compiler updater (flexdll bootstrap didn't used to work if you had flexlink in PATH with no flexdll.h). I can separate that, but I don't expect you're worried about that being here!
 - A "slower" version of Buffer.add_utf_16le_uchar is added to the compatibility layer for pre-4.06.0 builds
 - A very simple Uchar module is added to the compatibility layer for pre-4.03.0 builds (I haven't actually tested that, but I think from inspection that should be OK)

alainfrisch:
@dra27 Do you want to give a look at #46, or do you prefer me to create a release now?

dra27:
Sorry - I had, but I see that I was clearly expecting you to know my thoughts psychically! I'll go and commit them now...

PR #48, opened 2017-12-06
Allow linking C++

dra27:
At least for mingw64! These two changes together have allowed me to link applications using the ocaml-mccs C++ stubs.

I'm reasonably confident that the change to add `-lgcc_eh` is correct (because we already add `-lgcc`). I'm not so totally confident about the change in `Reloc.needed`, although I am confident that this only changes the behaviour for symbols which previously would have caused linker errors.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,6 +4,7 @@ Next version
 - GPR#49: Adding support for 64bit GNAT compiler (Johannes Kanig)
 - GPR#64: Fix problem with setting page permission during relocation (report
   by Dmitry Bely)
+- GPR#48: Enable linking of C++ object files for mingw ports (David Allsopp)
 
 Version 0.37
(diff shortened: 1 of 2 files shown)


dra27:
Buried in the quite sensible refactoring in https://github.com/cygwinports/flexdll/blob/master/0.34-cygwin.patch is a change in the Cygwin default libs which is related to this, as `-lgcc_s` is added.

@yselkowitz - why do you do `-lgcc_s -lgcc` rather than `-lgcc -lgcc_eh`? (or, put another way, if we add `-lgcc_eh`, would you still need to patch it to have `-lgcc_s` instead?)

yselkowitz:
While on other targets, `-lgcc_s` (aka `-shared-libgcc`) is limited to C++ usage, it is the default on Cygwin even with C (when not linking with `-static`, of course), in order to support throwing exceptions across DLL boundaries.

PR #49, opened 2018-01-09
Adding support for 64bit GNAT compiler

kanigsson:
This commit adds support for a 64bit variant of the GNAT compiler. It's basically the same as Mingw64, but using the already existing code to find the include PATH for GNAT.


alainfrisch:
Thanks!

PR #54, opened 2018-05-04
Fix some bugs, add some features

db4:
Fixes
- page access rights bug (as discussed in email conversation with Alain Frisch, Nov 25 2013)
- -custom-crt was broken w.r.t non-crt default libraries
- Ignore some x64 debug relocs that were previously missed

Features
- add some useful options
- support ECOFF format (generated by cl with /bigobj flag). No docs on ECOFF is available, so support was reverse-engineered from binutils sources (https://github.com/bminor/binutils-gdb)

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -2,6 +2,8 @@ Next version
 - GPR#58: Bug fix: fix -custom-crt regression (Dmitry Bely)
 - GPR#61: Ignore x64 debug relocs (Dmitry Bely)
 - GPR#49: Adding support for 64bit GNAT compiler (Johannes Kanig)
+- GPR#64: Fix problem with setting page permission during relocation (report
+  by Dmitry Bely)
 
 Version 0.37
 - Bug fix: the IMAGE_SCN_LNK_NRELOC_OVFL section flag was not propertly reset when
--- a/flexdll.c
+++ b/flexdll.c
@@ -146,25 +146,6 @@ static void dump_master_reloctbl(reloctbl **ptr) {
   while (*ptr) dump_reloctbl(*ptr++);
 }
 
-static void allow_write(char *begin, char *end, DWORD new, PDWORD old) {
-  static long int pagesize = 0;
-  int res;
-  SYSTEM_INFO si;
-
-  if (0 == pagesize) {
-    GetSystemInfo (&si);
-    pagesize = si.dwPageSize;
-  }
-
-  begin -= (size_t) begin % pagesize;
-  res = VirtualProtect(begin, end - begin, new, old);
-  if (0 == res) {
-    fprintf(stderr, "natdynlink: VirtualProtect failed (%s), begin = 0x%p, end = 0x%p, new = %x\n", ll_dlerror(), begin, end, new);
-    exit(2);
-  }
-  /* printf("%p -> %p\n", *old, new); */
-}
-
 /* Avoid the use of snprintf */
 static void cannot_resolve_msg(char *name) {
   static char msg[] = "Cannot resolve ";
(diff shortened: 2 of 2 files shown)


reloc.ml
```diff
@@ -898,6 +914,28 @@ let build_dll link_exe output_file files exts extra_args =
...
+
+    begin match !exports_files with
+      | [] -> ()
+      | files ->
+        let ht = Hashtbl.create 7 in
+        let exists s = try Hashtbl.find ht s with Not_found -> false in
+        List.iter
+          (fun file ->
+              List.iter
+                (fun name -> if not (exists name) then Hashtbl.add ht name true)
```
alainfrisch: This could just be: `(fun name -> Hashtbl.replace ht name ())`

flexdll_initer.c
```diff
@@ -24,8 +24,16 @@ extern int reloctbl;
...
+  char *s, buf[256];
+  DWORD len = GetEnvironmentVariable("FLEXDLL_RELOCATE", buf, sizeof(buf));
+  if (len > 0 && len <= sizeof(buf)) {
+    s = buf;
+  }
+  else {
+    /* fallback to the legacy method */
+    s = getenv("FLEXDLL_RELOCATE");
+    if (!s) { fprintf(stderr, "Cannot find FLEXDLL_RELOCATE\n"); return FALSE; }
+  }
```
dra27: Being somewhat pedantic, shouldn't this be the other way around - you start off using `getenv` and if that fails because `FLEXDLL_RELOCATE` is not set then you use the new behaviour and read `GetEnvironmentVariable`?

alainfrisch:
> as discussed in email conversation with Alain Frisch, Nov 25 2013

For future reference, and for interested readers, here is your original (excellent!) report:

````


Hi Alain,

I stumbled across an access violation bug caused by the way FlexDLL
performs relocations. When constructing dynamic relocation tables it
packs several consecutive relocation addresses into one "zone". When
performing actual relocations it allows write access via
VirtualProtect(zone_begin, zone_size, PAGE_EXECUTE_WRITECOPY) (BTW,
why not PAGE_EXECUTE_READWRITE?), then patches addresses and then
restore the old access rights via VirtualProtect(zone_begin,
zone_size, old_access). Unfortunately this is not entirely correct. I
see the situation where one zone covers 2 memory pages with different
access rights: PAGE_READONLY and PAGE_EXECUTE_READ. On restore the
second page will be set PAGE_READONLY and any attempt to execute code
from it will lead to an access violation error (as CPU supports NX and
DEP is enabled in the OS).

Relocation table that causes troubles:

Dynamic relocation table found at 634B13A0
 Non-writable relocation in zone 63483D57 -> 63483E83
 Non-writable relocation in zone 63483BB8 -> 63483CD5
 Non-writable relocation in zone 63480BE5 -> 634811B8
 634811B8 (kind:0100) (now:00C363E9)  caml_local_roots
 634811B4 (kind:0100) (now:00005E58)  caml_local_roots
 63480BE9 (kind:0100) (now:00000000)  caml_atom_table
 63480BE5 (kind:0100) (now:00000000)  caml_atom_table
 63483CD5 (kind:0102) (now:013D2900)  caml_copy_string
 63483CB8 (kind:0101) (now:0032D3F8)  w32_unicode_pack_to_string
 63483CA5 (kind:0102) (now:013FC224)  caml_atom_table
 63483C8C (kind:0101) (now:9DF62F20)  caml_initialize
 63483C4E (kind:0101) (now:9DF62BFE)  caml_alloc_shr
 63483BE9 (kind:0101) (now:0032D526)  w32_uuid_ml2c
 63483BB8 (kind:0101) (now:0032D4CB)  w32_uuid_c2ml
 63483E83 (kind:0102) (now:013FB4D0)  caml_local_roots
 63483E68 (kind:0102) (now:013FB4D0)  caml_local_roots
 63483D81 (kind:0102) (now:013FB4D0)  caml_local_roots
 63483D75 (kind:0102) (now:013FB4D0)  caml_local_roots
 63483D57 (kind:0102) (now:013FB4D0)  caml_local_roots

63480000 page was PAGE_READONLY
63481000 page was PAGE_EXECUTE_READ

Another problem is that one page can be covered by several zones. As
you restore access rights in the same order you modify them, they will
be incorrect in the end. I was able to fix this just replacing

  /* Restore permissions. Should do it also in case of failure... */
  for (wr = tbl->nonwr; wr->last != 0; wr++)
    allow_write(wr->first,wr->last + 4,wr->old,&wr->old);

with

  /* Restore permissions in the reverse order. Should do it also in
case of failure... */
  wr = tbl->nonwr;
  while ((wr+1)->last != 0)
    wr++;
  for (; wr >= tbl->nonwr; wr--)
    allow_write(wr->first,wr->last + 4,wr->old,&wr->old);

but I still have no idea how to fix the first problem. Now I just use

    allow_write(wr->first,wr->last + 4,(wr->old ==
PAGE_READONLY)?PAGE_EXECUTE_READ:wr->old,&wr->old);

as a workaround in hope that someone will fix it properly.
````

and my proposal fix strategy at that time:

````
1. First version:  one changes the right accesses for each relocation and immediately restore it after the patch is done.  This even simplifies the work of flexlink.exe, since it no longer needs to compute these "zones".

2. Second version: *if* it turns out to be too expensive to change the rights on memory pages for each relocation, one can implement the simplest strategy of checking if the next relocation falls in the same page as the previous one, and in this case, avoid the two useless calls to VirtualProtect.  One should then arrange so that flexlink.exe emits relocations in a sorted order (it might already be the case). 
````

I think one should evaluate the performance impact of changing access rights on each relocation.  There can be many relocations, and I'm concerned that this could introduce a significant slowdown when loading a DLL.  Have you tried to assess the performance impact?

**Comment by dra27** [comment]
The `putenv` thing is surprising me, because `putenv` **does** call `SetEnvironmentVariable` (see the CRT sources). Do you have a repro case? I'm surprised, given where flexdll hooks the startup, that both CRTs are able to load *and cache the environment* in time for this to be a problem, but I guess I'm just not seeing it?

File: reloc.ml (hunk 1 above)
Comment (db4) [reply reloc.ml]: Indeed; sometimes old code looks really strange :) Will fix that.

File: flexdll_initer.c (hunk 2 above)
Comment (db4) [reply flexdll_initer.c]: I don't see much difference but if you think it's important I can do it your way.

Comment (dra27) [reply flexdll_initer.c]: Just a highly pedantic "do it the old way and only if that fails do it the new way"! See my other comment about the `putenv` stuff - I'm slightly confused by it and curious to see the failure. The issue with putenv/getenv will not be that putenv doesn't call `SetEnvironmentVariable`, it's that getenv cached `GetEnvironmentVariable` (or rather `GetEnvironmentStrings` or whatever it's called)

Comment (dra27) [reply flexdll_initer.c]: I was trying to work out if a program loading a flexdll DLL which also loads a flexdll DLL could fail - but that shouldn't, because it's always the same flexdll_relocate you want!

**Comment by db4** [comment]
> The putenv thing is surprising me, because `putenv` does call `SetEnvironmentVariable` (see the CRT sources).

Yes, you are right. I can't reproduce the problem with modern Visual Studio so I will exclude the questionable commit.

File: reloc.ml [hunk 3]
```diff
@@ -389,8 +389,11 @@ let add_reloc_table obj obj_name p =
         | `x64, 0x08 when not !no_rel_relocs-> 0x0003 (* rel32_4 *)
         | `x64, 0x06 when not !no_rel_relocs-> 0x0005 (* rel32_2 *)
 
-        | `x86, (0x0a (* IMAGE_REL_I386_SECTION *) |
-                 0x0b (* IMAGE_REL_I386_SECREL*) ) ->
+        | `x64, 0x03 (* IMAGE_REL_AMD64_ADDR32NB *) ->
+            0x0100 (*  Win64 EH data structures relocs: ignore *)
```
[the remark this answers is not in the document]
db4: Well, if we don't ignore them, `flexlink` fails to process x64 .obj files that were compiled by C++ compiler. If we ignore them, we just leave them 'as is' to Win32 loader, not breaking anything. Maybe to add some tests to be sure that it really works?

db4:
> IIRC having mixed static and shared CRT is not recommended at all, though?

Well, you can model the same situation using shared CRTs of different Visual Studio versions. It's not uncommon when different DLLs in one application are linked with different CRTs. Not very likely for OCaml plugins, but flexdll is advertised as a general-purpose tool, right?

> That said, it's all academic for flexdll - the importing thing is communicating that function pointer, which at some point we'll do using something better than an environment variable anyway!

Several years ago I hit the problem and fixed it the most simple way. If we introduce more reliable method of passing a pointer - of course, it would be better. But I doubt it will happen in the foreseeable future :)

reloc.ml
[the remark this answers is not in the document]
alainfrisch: > If we ignore them, we just leave them 'as is' to Win32 loader, not breaking anything.

Well, if we ignore them, we are actually removing them from the locations managed by Win32 and we won't patch the location in our loader either.  Meaning that the memory address could be left with a dummy value, possibly resulting in arbitrary behavior when the value is used at runtime.

alainfrisch:
Any thought on:

> I think one should evaluate the performance impact of changing access rights on each relocation. There can be many relocations, and I'm concerned that this could introduce a significant slowdown when loading a DLL. Have you tried to assess the performance impact?

reloc.ml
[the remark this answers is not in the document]
db4: I see. This relocation kind is RVA (relative virtual address), and it's easy to handle it correctly. I will try to do that. Unfortunately flexdll does not contain comprehensive relocation tests to add some test case to. How to demonstrate that this kind really works under flexdll/x64?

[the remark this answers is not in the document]
alainfrisch: I would first create a test case that fails at link time with the current version and at runtime when ignoring the relocation, and check that it works fine after adding proper support for the relocation.

What about splitting the PR into multiple ones?  I understand this is more work on your side, but it will simplify discussing and merging rather independent parts.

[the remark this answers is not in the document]
db4: > I would first create a test case that fails at link time with the current version and at runtime when ignoring the relocation, and check that it works fine after adding proper support for the relocation.

Ok

> What about splitting the PR into multiple ones?

I can create separate PRs for all six commits, no problem.

PR #58, opened 2018-07-25
-custom-crt regression

db4:
-custom-crt was broken w.r.t non-crt default libraries


alainfrisch:
The regression was introduced in 3fa3b15f327628ba45fd30a00d1136c043e69b73, right?

db4:
Yes, looks like that.

PR #59, opened 2018-07-25

db4:
Support ECOFF format (generated by cl with /bigobj flag). No docs on ECOFF is available, so support was reverse-engineered from binutils sources (https://github.com/bminor/binutils-gdb)

### DIFF
--- a/coff.ml
+++ b/coff.ml
@@ -162,6 +162,7 @@ and section = {
 }
 
 type coff = {
+  bigobj: bool;
   obj_name: string;
   machine: int;
   date: int32;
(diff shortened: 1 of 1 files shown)


alainfrisch:
The PR shouldn't affect the non-ECOFF case, and I assume you have tested the ECOFF support (which I know nothing about), so I'm going to  merge!

alainfrisch:
Sorry, I was too fast.  This seems to break with 4.06 (see AppVeyor).  Can you look at it?

alainfrisch:
I could not reproduce the problem locally.  But I wonder about the change in Lib.read_lib.  Can you explain why we need to check two extra bytes (= 0) to decide between import vs object?

PR #63, opened 2018-07-31

alainfrisch:
This is #59 from @db4 again, which I merged and then reverted.  Waiting for AppVeyor feedback.

### DIFF
--- a/coff.ml
+++ b/coff.ml
@@ -162,6 +162,7 @@ and section = {
 }
 
 type coff = {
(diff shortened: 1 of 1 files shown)


db4:
I've managed to identify the problem. Probably it's better to close this PR so I could create a new one?

alainfrisch:
Yes, you can continue working on your branch and open an PR from it, I'll close the current PR.

PR #64, opened 2018-08-01
Fix problem with setting page permission during relocation

alainfrisch:
Another take at #57, avoiding repeated calls to VirtualProtect on the same pages for performance reasons.

Loading the big .cmxs I mentioned in #57 now takes about 0.22s (against 0.20s before the change, and against 0.65s with the version in #57).  One could try to be even more clever and group relocation that fall in adjacent pages, but this becomes a bit more tricky.


db4:
Looks good!

alainfrisch:
Following this PR, one could avoid computing those 'nonwr' stuff.  Cf https://github.com/alainfrisch/flexdll/commit/47e7180eeca6b2c4856964ed5541543da57c4c3f

This breaks compatibility between versions of flexdll, for very little gain.  So perhaps it's not worth doing it now.

PR #65, opened 2018-08-03

db4:
Looks like it works now.  Actually, ECOFF is not very different from COFF: it uses another file header (ANON_OBJECT_HEADER_BIGOBJ, see winnt.h in Windows SDK) and allows more than 2^16 sections in a single file (`NumberOfSections` header field. has tipe `DWORD`)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -207,7 +207,7 @@ flexdll_initer_mingw64.o: flexdll_initer.c
 
 
 demo_msvc: flexlink.exe flexdll_msvc.obj flexdll_initer_msvc.obj
-	$(MSVC_PREFIX) $(MAKE) -C test clean demo CHAIN=msvc CC="$(MSVCC)" O=obj
+	$(MSVC_PREFIX) $(MAKE) -C test clean demo CHAIN=msvc CC="$(MSVCC)" BIGOBJ="/bigobj" O=obj
 
 demo_cygwin: flexlink.exe flexdll_cygwin.o flexdll_initer_cygwin.o
 	$(MAKE) -C test clean demo CHAIN=cygwin CC="$(CYGCC)" O=o
@@ -222,7 +222,7 @@ demo_mingw64: flexlink.exe flexdll_mingw64.o flexdll_initer_mingw64.o
 	$(MAKE) -C test clean demo CHAIN=mingw64 CC="$(MIN64CC)" O=o
 
 demo_msvc64:  flexlink.exe flexdll_msvc64.obj flexdll_initer_msvc64.obj
-	$(MSVC64_PREFIX) $(MAKE) -C test clean demo CHAIN=msvc64 CC="$(MSVCC64)" O=obj
+	$(MSVC64_PREFIX) $(MAKE) -C test clean demo CHAIN=msvc64 CC="$(MSVCC64)" BIGOBJ="/bigobj" O=obj
 
 distclean: clean
 	rm -f Makefile.winsdk
--- a/coff.ml
+++ b/coff.ml
@@ -901,11 +901,15 @@ module Lib = struct
     let obj size name =
 (*      Printf.printf "-> %s (size %i)\n" name size;  *)
       let pos = pos_in ic in
-      if size > 18 && read_str ic pos 6 = "\000\000\255\255\000\000"
-      then imports := Import.read ic pos size :: !imports
-      else objects := (name,
-                       Coff.get ic pos pos
-                         (Printf.sprintf "%s(%s)" libname name)) :: !objects
+      let buf = read ic pos 6 in
+      let coff = int16 buf 0 != 0 || int16 buf 2 != 0xFFFF in
+      let version = int16 buf 4 in
+      if size > 18 && not coff && version = 0 then
+        imports := Import.read ic pos size :: !imports
+      else if coff || (not coff && version >= 2) then
+        objects := (name,
+                    Coff.get ic pos pos
+                      (Printf.sprintf "%s(%s)" libname name)) :: !objects
     in
     let rec read_member () =
       let buf = read ic (pos_in ic) 60 in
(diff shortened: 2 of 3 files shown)


bschommer:
What is the status on this PR? Having support for `/bigobj` would be a nice to have feature.

dra27:
Ah: somehow I'd seen the first bit, and not noticed that it was plumbed in in `test/Makefile` too! This might just be related to another issue we've just started seeing with the latest version of the Windows SDK, but let's not get hopes up too much. I clearly do have less excuse to delay reviewing, beyond needing to refactor to test both with and without the switch on msvc...!

coff.ml
```diff
@@ -776,17 +846,31 @@ module Coff = struct
...
         (fun s -> incr no; assert(s.sec_pos < 0); s.sec_pos <- !no)
         x.sections
     in
 
-    emit_int16 oc (List.length x.sections);
-    emit_int32 oc x.date;
+    if x.bigobj then begin
+      emit_int16 oc 0; (* Sig1 *)
+      emit_int16 oc 0xffff; (* Sig2 *)
+      emit_int16 oc 2; (* Version *)
```
dra27, line 859: Why is this written on this side (for bigobj) but not on the other (for the legacy one)?

There may well be a reason, I was just scratching my head trying to work it out!

coff.ml
```diff
@@ -859,11 +946,17 @@ module Lib = struct
     let obj size name =
 (*      Printf.printf "-> %s (size %i)\n" name size;  *)
       let pos = pos_in ic in
-      if size > 18 && read_str ic pos 4 = "\000\000\255\255"
-      then imports := Import.read ic pos size :: !imports
```
dra27, line 863: Was this an error before - having ascertained that the magic header was there, shouldn't have read the version, instead doesn't it get lumped onto the first symbol?

dra27 (commented):
There were some minor refactorings - mainly to have slightly fewer names for numbers (I'd mis-characterised what I'd got before).

There are a couple of questions about the changes to the header reading.

For testing - does the current test definitely exercise the code which _writes_ the new ECOFF file and is it then read back by another part of the test? (i.e. do we effectively do a round-trip of the file somewhere in amongst!).

Assuming that is the case, I think all that needs to happen is to do one test on MSVC with `/bigobj` and another without, which should be an easy tweak.

Because of #94, flexdll 0.40 will be released soon - as long as I can convince myself that the changes here definitely don't affect the normal COFF case (that's the basis for the questions about the version number reading following the `\000\000\255\255` header) then I think this can (finally) be included, thanks!

coff.ml
db4: I address your other comments later (they're mostly cosmetic and I have nothing against them. I tried to change the original sources minimally while you're trying to make them more readable). Regarding this:
 https://github.com/alainfrisch/flexdll/blob/f40ed2d57bbb70ac8223d30a6ffa737f2c0e430e/coff.ml#L856-L873
I don't quite understand your question. Why Sig1, Sig2, and Version are not emitted for the standard COFF? Just because its header does not include these fields.

coff.ml
db4: Sort of. This check would treat ANON_OBJECT_HEADER_BIGOBJ as IMPORT_OBJECT_HEADER (i.e. an import library). I didn't try to feed bigobj files to the current version of flexlink but suppose it will fail somewhere.

dra27: Ah, I see now (sorry, I could also have scoured the COFF spec, too) - what we had before was an assumption that the `\000\000\255\255` header implied strictly an object library whereas now we actually read the header properly. Gotcha!

coff.ml
dra27: This makes sense to me now w.r.t. the comment below - I hadn't registered properly that what we had before was an incorrect reading of an ECOFF header (assuming an import library)

coff.ml
```diff
@@ -397,7 +402,12 @@ module Symbol = struct
             "value=0x%08lx, sect=%s, storage=%s, aux=%S\n"
             s.value sect storage (Bytes.to_string s.auxs)
 
-  let put strtbl oc s =
+  let put bigobj strtbl oc s =
```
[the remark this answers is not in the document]
db4: This breaks
https://github.com/alainfrisch/flexdll/blob/f40ed2d57bbb70ac8223d30a6ffa737f2c0e430e/coff.ml#L885

dra27:
You'll also want something like https://github.com/dra27/flexdll/commit/31e835f853eaabb4194b6cc4e6ac0bf7ffa65701, I'm afraid!

dra27:
Fortunately, once the AppVeyor build of master has finished the caches will all be primed, so the next build should be quite a bit quicker! We've (at OCaml Labs) nearly got our Windows build cluster up and running... I'm hoping to replace the AppVeyor CI with a rather more parallel one on that!

db4:
@dra27 I think I've addressed all your comments (except labeling `bigobj` that complicates partial application). CI is all green :)

Makefile
```diff
@@ -207,7 +207,7 @@ flexdll_initer_mingw64.o: flexdll_initer.c
 
 
 demo_msvc: flexlink.exe flexdll_msvc.obj flexdll_initer_msvc.obj
-	$(MSVC_PREFIX) $(MAKE) -C test clean demo CHAIN=msvc CC="$(MSVCC)" O=obj
+	$(MSVC_PREFIX) $(MAKE) -C test clean demo CHAIN=msvc CC="$(MSVCC)" BIGOBJ="/bigobj" O=obj
```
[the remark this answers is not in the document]
db4: Ahh, I just forgot that I did 3 years ago :( My last commit adding
https://github.com/alainfrisch/flexdll/blob/4da3dac05d715f25a27d7b56bc37d9dde09e37f6/appveyor_build.sh#L168-L170
is useless, /bigobj test was alreary there (added in 7ccf31b99d30efeb27c0cd4aacf7765f9d733845). Will drop the last commit instead.

Issue #72, opened 2019-04-28
Support gcc "-Wl" linker pass through option


MSoegtropIMC:
This is a follow up on the discussion on Coq PR 9729:

<https://github.com/coq/coq/pull/9279>

If flexlink would support -Wl as synonym for -link, libraries which use -Wl options could be compiled without changes in the build instructions on Windows.

I will provide a PR, but only for simple cases of the -Wl syntax.

dra27:
This issue should now be fixed, with the more complex case addressed in #85

PR #73, opened 2019-04-28
Implement issue #72 "Support gcc "-Wl" linker pass through option"

MSoegtropIMC:
This PR implements issue #72.

### DIFF
--- a/cmdline.ml
+++ b/cmdline.ml
@@ -56,6 +56,7 @@ let footer =
 "Notes:
 * The -I, -l and -L options do not need to be separated from their argument.
 * An option like /linkXXX is an abbrevation for '-link XXX'.
+* An option like -Wl,-XXX is an abbreviation for '-link -XXX'.
 * FlexDLL's object files are searched by default in the same directory as
   flexlink, or in the directory given by the environment variable FLEXDIR
   if it is defined.
--- a/cmdline.ml
+++ b/cmdline.ml
@@ -249,6 +249,9 @@ let parse_cmdline () =
         String.sub s 0 2 :: String.sub s 2 (String.length s - 2) :: tr rest
     | s :: rest when String.length s >= 5 && String.sub s 0 5 = "/link" ->
         "-link" :: String.sub s 5 (String.length s - 5) :: tr rest
+    (* Convert gcc linker option prefix -Wl, to flexlink linker prefix -link *)
+    | s :: rest when String.length s >= 6 && String.sub s 0 5 = "-Wl,-" ->
+        "-link" :: String.sub s 4 (String.length s - 4) :: tr rest
     | "-arg" :: x :: rest ->
         tr (Array.to_list (Arg.read_arg x)) @ rest
     | "-arg0" :: x :: rest ->


yakobowski:
This seems very desirable indeed.

MSoegtropIMC:
I somehow lost track of this PR - need to check why PR fails here. We are using this patch in Coq since quite a while and it works fine there.

MSoegtropIMC:
The CI failure is due to incompatibility with OCaml master. It is the same failure as all other recent tests have. See e.g.
<https://ci.appveyor.com/project/alainfrisch/flexdll/build/1.0.90>
So from my point of view this PR is good for integration.

dra27:
This is a good idea - would you have a chance to rebase it (CI should then pass - the failure was indeed unrelated), but could you also a note about `-Wl,` options being passed to the linker in `Cmdliner.footer`.

I think it's acceptable for these to be passed on verbatim to the MSVC linker as well - in cases where it's a common option it'll work, and we can deal with actually translating options as and when the need arises.

MSoegtropIMC:
Sure I will rebase and look and look into -Wl

dra27:
I took the liberty of rebasing this - @alainfrisch, does it look OK to you?

MSoegtropIMC:
> I took the liberty of rebasing this - @alainfrisch, does it look OK to you?

Thanks - I appreciate it!

Your comment means that I should update the documentation?

Sorry for the delays - I am quite busy these days.

dra27:
@MSoegtropIMC - that's no problem at all! The comment I'd left was to mention that `-Wl,` is supported when you invoke `flexlink -help` which I've done in https://github.com/alainfrisch/flexdll/pull/73/commits/20f0d493d74ce6c78172fe4bf9d305a081426272

MSoegtropIMC:
> which I've done in 20f0d49

Thanks! One might want to mention that one can pass only one argument with -Wl at a time - several comma separated arguments as the syntax might suggest and gcc supports it - are not supported. Or I (or someone) could fix this ...

MSoegtropIMC:
P.S.: One can give of cause several -Wl arguments.

dra27:
Ah, that's a good point - is that simply a matter of splitting on `,` characters. or does one have to be more intelligent than that w.r.t. any embedded strings (e.g. is `-Wl,-foo="a,b",-bar` also legal?)

dra27:
(my impression from https://gcc.gnu.org/onlinedocs/gcc/Link-Options.html#Link-Options is that the splitting is simply done on commas)

MSoegtropIMC:
> Ah, that's a good point - is that simply a matter of splitting on `,` characters. or does one have to be more intelligent than that w.r.t. any embedded strings (e.g. is `-Wl,-foo="a,b",-bar` also legal?)

I would think that the quotes are handled by the shell anyway, so that flexlink doesn't see them. So my assumption is that a simple split on , is sufficient. The only question is what happens with double commas or commas at the end, but I think once can do this garbage in - garbage out. The point of this is compatibility with make files targeting gcc and it is sufficient to support what is out there in the wild.

alainfrisch:
> I took the liberty of rebasing this - @alainfrisch, does it look OK to you?

Yes, this LGTM!  (Perhaps just update the CHANGES file?)

dra27:
If you're happy to push a commit with the `String.split`, I'll update the compatibility layer so that 4.03 and earlier don't complain!

(noted re changes, @alainfrisch - although I was going to go through the branch history and update it shortly too - I think we're nearly at a collected enough point for 0.38!)

dra27:
Actually, the compatibility layer change isn't entirely trivial, but I think that the patch is useful to you even when it can online handle a single argument, right? So I've merged this and put my WIP for the splitting into #85.

Issue #51, opened 2018-02-02
Option -custom-crt should not add MSVC linker option /nodefaultlib:LIBCMT


ccuchet:
When one wants to use LIBCMT.lib in place of default CRT library MSVCRT.lib, it is not possible with FlexDLL v0.35, because option -custom-crt prevents both to use MSVCRT.lib and LIBCMT.lib.
It should only disable usage of MSVCRT.lib.

See reloc.ml lines 1005-1006:
  if !custom_crt then "/nodefaultlib:LIBCMT /nodefaultlib:MSVCRT " ^ extra_args
  else "msvcrt.lib " ^ extra_args
should be:
   if !custom_crt then "/nodefaultlib:MSVCRT " ^ extra_args
   else "msvcrt.lib " ^ extra_args
 
Thanks

dra27:
Can I ask what your use-case is here? (welcome to GitHub, by the way!)

@alainfrisch - why was `-custom-crt` added? Of possible relevance, in case your memory is as bad as mine at the moment: https://sympa.inria.fr/sympa/arc/caml-list/2009-07/msg00024.html.

As it happens, I think the fix should possibly go further and not specify either `/nodefaultlib`, since the reason msvcrt.lib would be pulled in anyway would be because the user was linking objects compiled with `cl /MD`, so it should really be the caller's responsibility to say `-custom-crt` (disable flexlink automatically linking with the runtime dll) and then `-link /nodefaultlib:msvcrt` (prevent the linker automatically adding it in, because I'm linking objects which expected it).

ccuchet:
Hello

We would like to use LIBCMT and not MSVCRT in an OCaml application.
We compile C files with option '/MT' (not '/MD) and would like to link with LIBCMT.lib (and not MSVCRT.lib).

It is fine if FlexDLL option '-custom-crt' just avoids adding 'msvcrt.lib' and does nothing more.

Thanks

alainfrisch:
Yes, my memory is at least as bad as yours.  I think -custom-crt was an experiment to allow linking with libcmt, but I never really managed to make it work full scale.

@ccuchet After the suggested fix (or the "going further" variant), are you indeed able to link your OCaml application with libcmt?  You probably need to recompile at least the OCaml runtime system, no?

ccuchet:
Yes we compile the OCaml runtime with option /MT 
and after the suggested fix we can link with no errors with LIBCMT.lib
(we do not use dynamic loading of DLL in our application).

[...]

PR #74, opened 2019-05-28
make it that -custom-crt does not prevent linking with libcmt.lib

oandrieu:
*c.f.* #51

This does removes the addition of the `/nodefaultlib:LIBCMT` in the command line when using the `-custom-crt` option.
With this, we can successfully link an OCaml program with LIBCMT by invoking ocamlopt with `-ccopt -custom-crt -cclib libcmt.lib` (with a suitably compiled runtime library).

fixes issue #51

### DIFF
--- a/reloc.ml
+++ b/reloc.ml
@@ -1002,7 +1002,7 @@ let build_dll link_exe output_file files exts extra_args =
         let c = open_out implib in output_string c "x"; close_out c;
         let _impexp = add_temp (Filename.chop_suffix implib ".lib" ^ ".exp") in
         let extra_args =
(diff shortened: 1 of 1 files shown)


alainfrisch:
But wouldn't it prevent linkking with other variants of the C runtime?  I would have expected that /nodefaultlib:LIBCMT doesn't hurt if libcmt.lib is passed explicitly.  Is it an ordering problem between libraries?

oandrieu:
Err yeah indeed, `/nodefaultlib:LIBCMT` doesn't hurt. I'm not quite sure what Christophe's issue was about then.
Nevermind, sorry for the noise Alain.

PR #78, opened 2019-10-15
Added compatibility layer for newer ocaml.

bschommer:
Pervasives is deprecated since ocaml 4.08 and should be replaced
by Stdlib.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -138,7 +138,7 @@ COMPILER-$(OCAML_VERSION):
 
 test_ver = $(shell if [ $(OCAML_VERSION) -lt $(1) ] ; then echo lt ; fi)
 
-Compat.ml: COMPILER-$(OCAML_VERSION) $(if $(call test_ver,4020),Compat402.ml) $(if $(call test_ver,4030),Compat403.ml) $(if $(call test_ver,4050),Compat405.ml) $(if $(call test_ver,4060),Compat406.ml)
+Compat.ml: COMPILER-$(OCAML_VERSION) $(if $(call test_ver,4020),Compat402.ml) $(if $(call test_ver,4030),Compat403.ml) $(if $(call test_ver,4050),Compat405.ml) $(if $(call test_ver,4060),Compat406.ml) $(if $(call test_ver,4070),Compat407.ml)
 	cat $^ > $@
 
 flexlink.exe: $(OBJS) $(RES)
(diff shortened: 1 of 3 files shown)


Compat408.ml
```diff
@@ -0,0 +1,11 @@
...
+(*   FlexDLL                                                            *)
+(*   Alain Frisch                                                       *)
+(*                                                                      *)
+(*   Copyright 2007 Institut National de Recherche en Informatique et   *)
+(*   en Automatique.                                                    *)
+(************************************************************************)
+
+module Stdlib = struct
+  let compare = Pervasives.compare
+end
```
dra27, line 11: Can't this just be...
```suggestion
module Stdlib = Pervasives
```
... and then any future uses of `Stdlib.` should be safe(ish) too?

Compat408.ml
```diff
@@ -0,0 +1,11 @@
+(************************************************************************)
```
dra27, line 1: This file should be `Compat407.ml`

Makefile
```diff
@@ -138,7 +138,7 @@ COMPILER-$(OCAML_VERSION):
 
 test_ver = $(shell if [ $(OCAML_VERSION) -lt $(1) ] ; then echo lt ; fi)
 
-Compat.ml: COMPILER-$(OCAML_VERSION) $(if $(call test_ver,4020),Compat402.ml) $(if $(call test_ver,4030),Compat403.ml) $(if $(call test_ver,4050),Compat405.ml) $(if $(call test_ver,4060),Compat406.ml)
+Compat.ml: COMPILER-$(OCAML_VERSION) $(if $(call test_ver,4020),Compat402.ml) $(if $(call test_ver,4030),Compat403.ml) $(if $(call test_ver,4050),Compat405.ml) $(if $(call test_ver,4060),Compat406.ml) $(if $(call test_ver,4080),Compat408.ml)
```
dra27, line 141: `Stdlib` was introduced in 4.07.0 - it's the deprecation which is specific to 4.08.0, so the test should be for `4070` and the file added `Compat407.ml`

Compat408.ml
bschommer: I change as suggested, original I just wanted to make sure that only the function needed are imported, but you are right it might be more useful to directly import the whole module.

Compat408.ml
bschommer: Is renamed.

Makefile
bschommer: Should be changed to the new name.

dra27:
This looks good to go, thanks - would you be able to rebase and squash just so we get a full CI run?

bschommer:
I just rebased and squashed.

dra27:
Great, thanks!

PR #80, opened 2020-04-11
Ensure commands always silent for MSVC

dra27:
If a command is routed through bash because it's too long, the fact it's meant to be piped to NUL is forgotten. Ensure the bash command is piped to >/dev/null (with the slightly tedious recovery of the script needed if the command fails).

#68 needs to be merged first, because it corrects a magic number in this PR

### DIFF
--- a/reloc.ml
+++ b/reloc.ml
@@ -80,7 +80,7 @@ let max_command_length =
 let get_output ?(use_bash = false) ?(accept_error=false) cmd =
   let fn = Filename.temp_file "flexdll" "" in
   let cmd' = cmd ^ " > " ^ (Filename.quote fn) in
-    if String.length cmd' < max_command_length && not use_bash then
+    if String.length cmd' <= max_command_length && not use_bash then
       begin
         if (Sys.command cmd' <> 0) && not accept_error
         then failwith ("Cannot run " ^ cmd);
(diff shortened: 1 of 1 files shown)


reloc.ml
```diff
@@ -198,27 +198,35 @@ let new_cmdline () =
...
 
     if !verbose >= 1 then Printf.printf "(call with bash: %s)\n%!" fn;
-    if Sys.command (Printf.sprintf "bash %s" fn) <> 0 then
+    let invoke = Printf.sprintf "bash %s" fn in
+    if Sys.command invoke <> 0 then begin
+      if pipe_to_null then begin
+        let oc = open_out fn in
+        output_string oc cmd;
+        close_out oc;
+        ignore (Sys.command invoke)
```
alainfrisch, line 223: What's the purpose of running the command again without the redirection?  We don't do that in the "direct" case (where bash is not used); and we might get errors (printed to stderr) twice.

dra27: Yes we do - see deleted lines 220 and 221 below.

It's a good point, though - the treatment on the other branch is for MSVC where errors are also sent to stdout. I think the correct thing is probably to do `&>/dev/null` on this side and a belt-and-braces `2>&1` on the other?

dra27: The whole lot exists to get rid of those pesky "foo.c" lines from MSVC...

dra27:
Rebased and I think ready to merge?

PR #81, opened 2020-04-11
Switch from findwinsdk to msvs-detect

dra27:
This extracts part of #17 (which I still want to come back to). `findwinsdk` is an early version of what became `msvs-detect` which now lives at https://github.com/metastack/msvs-tools.

The benefits of the switch are threefold:
- It's slightly faster, inasmuch as it detects the x86 and x64 compiler in a single call
- It's maintained, so it detects more recent Visual Studio installations
- It fixes AppVeyor (or at least it did on the run on my fork...)


dra27:
I intentionally haven't updated the AppVeyor matrix yet, pending #78

appveyor.yml
```diff
@@ -10,10 +10,10 @@ environment:
     CYG_CACHE: C:/cygwin/var/cache/setup
   matrix:
     - OCAMLBRANCH: 4.03
+      SKIP_OCAML_TEST: no
```
dra27, line 13: Every PR should ensure that the oldest bootstrap works

PR #86, opened 2020-06-26
Restore building on very old OCamls

dra27:
I intend to overhaul this to switch to GitHub Actions, as we'll get much better parallelism, but in the meantime.

I do not propose that we actively support building `flexlink` on OCaml 3.11, but it is quite useful that it basically still does (especially when opam finally allows installing the old Windows compilers in switches). I propose that dropping support for building flexlink with very old OCamls should be done consciously as part of a larger - long overdue - overhaul of the code, not just because of little patches.

With that in mind:
- restoring 4.02 support involves a slight hack to the Compat system. As part of #85, I will switch this to be rather better pre-processed, but for now it just involves moving `Buffer.to_bytes` into the Compat406 and not caring that it gets redefined for OCaml 4.03-4.05 when it doesn't need to be
- Restoring 3.11-4.01 support involves converting a single match on an `exception` back to `try`

The AppVeyor matrix is then expanded - I debated getting it to skip most versions for PRs, but I propose just putting this in as-is for now, given that the traffic on this repo I so low...

### DIFF
--- a/Compat401.ml
+++ b/Compat401.ml
@@ -6,15 +6,5 @@
 (*   en Automatique.                                                    *)
 (************************************************************************)
 
-(* Back-port required functionality from Bytes in 4.02.0 *)
-type bytes = string
-module Bytes = struct
-  include String
-
-  let blit_string = blit
-  let sub_string = sub
-  let of_string x = x
-  let to_string x = x
-  let cat = (^)
-end
-let output_bytes = output_string
+(* Introduced in 4.01.0 *)
+let ( |> ) x f = f x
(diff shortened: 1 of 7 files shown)


dra27:
Note that while the first time run is very slow for this, caching makes the AppVeyor run much quicker (see [this run](https://ci.appveyor.com/project/dra27/flexdll/builds/33759713) on my fork). As far as I can tell, when I (or someone) switches this to a well-engineered GitHub Actions workflow, we'd get a lot of parallelism for free.

PR #88, opened 2020-07-08
Be more forgiving with relative paths in search dirs

yakobowski:
If a search dir is relative and starts by '..', the current code is going
to crash. Instead, accept it without normalizing it.

### DIFF
--- a/reloc.ml
+++ b/reloc.ml
@@ -1217,7 +1217,11 @@ let normalize_path path =
   let rec loop acc path =
     match path with
     | "."  :: path -> loop acc path
-    | ".." :: path -> loop (List.tl acc) path
+    | ".." :: path -> begin
+        match acc with
+        | ".." :: _ | [] -> loop (".." :: acc) path
+        | _ :: acc -> loop acc path
+      end
     | elem :: path -> loop (elem :: acc) path
     | [] -> List.rev acc
   in


reloc.ml
```diff
@@ -1203,7 +1203,11 @@ let normalize_path path =
   let rec loop acc path =
     match path with
     | "."  :: path -> loop acc path
-    | ".." :: path -> loop (List.tl acc) path
+    | ".." :: path -> begin
+        match acc with
+        | _ :: q -> loop q path
+        | [] -> loop [".."] path
```
dra27, line 1209: This does prevent the crash, but it doesn't always return the right result - e.g. `../..` would normalise to `[]` instead of `[".."; ".."]`. I think this tweak fixes it:
```suggestion
        | previous :: acc when previous <> ".." -> loop acc path
        | _ -> loop (".." :: acc) path
```

dra27 (commented):
Sorry for the slow response time. Out of curiosity, have you actually seen this crashing?

reloc.ml
yakobowski: Duh 🤕 . Thanks for catching this!

yakobowski:
To answer your question, yes we saw this crashing. However we are building in a very specific env:
- regular MinGW compiler (i.e. not the "cross" one from Cygwin)
- tons of systems libraries at non-standard paths.
So I cannot extract a repro case, I'm afraid.

yakobowski:
I rebased and pushed a minor variation which does not use `when`. Apologies for the back-and-forth.

dra27:
Thank you for swift update! That looks great to me. No problem about the repro - I was only curious as to whether you'd got a "real" setup which hit this, or if you'd just noticed it reading the code.

I shall shortly tag 0.40, which will include this.

PR #89, opened 2020-09-17
Stop passing --image-base on Cygwin64

dra27:
On Cygwin64, `--image-base 0x10000` is always passed to the linker. This generates invalid DLLs for Cygwin64. This has never worked reliably, as DLLs need unique base addresses on Cygwin to survive fork. However, since version 4.4.4 Cygwin's rebase forces DLLs to the correct base address in 0x2:00000000-0x4:00000000. Since Cygwin64 executables load in 0x1:00000000-0x1:80000000, DLLs are always too far RELOC_REL32.

Rather than dynamically generating branch islands, the new approach on the OCaml-side is to return to using `__declspec(dllimport)` only, which will cause the linker to do that work for us, as it should.

Regardless of the solution, `--image-base` should cease being passed to the linker on Cygwin64.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,4 +1,6 @@
 Next version
+- GPR#89: Stop passing --image-base on Cygwin64 - Cygwin64 DLLs should load at
+  0x4:00000000-0x6:00000000 if not rebased or 0x2:00000000-0x4:00000000 if rebased. (David Allsopp)
 - GPR#90: Fix passing NULL to flexdll_dlopen on non-Cygwin builds (David Allsopp)
 
 Version 0.38
--- a/appveyor_build.sh
+++ b/appveyor_build.sh
@@ -164,6 +164,8 @@ for CHAIN in $CHAINS; do
 done
 
 for CHAIN in $CHAINS; do
+    # Decision on operation of Cygwin64 not yet taken
+    if [ "$CHAIN" = "cygwin64" ] ; then continue; fi
     run "make demo_$CHAIN" make demo_$CHAIN
 done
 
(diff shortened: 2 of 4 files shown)


dra27:
@alainfrisch - what do you think about this one? It will completely break shared library compilation of OCaml < 4.12 for Cygwin64, but equally it is broken at the moment, just more subtly (`fork` doesn't work). Equally, we should not be compiling DLLs for Cygwin64 at all like this. Cygwin's `rebaseall` can "corrupt" them at any moment by changing their base address - I think it only works at the moment because Cygwin Setup's rebase operation looks only at new DLLs.

An alternative would be to add a new flag which disables this automatic passing of `--image-base`, but it feels odd to have an explicit flag in order to select _correct_ behaviour over incorrect behaviour?!

alainfrisch:
Yes, I agree with your plan.  Breaking "more" OCaml < 4.12 for Cygwin64 is ok.

dra27:
Ta - I'll cut a release in the next couple of days and ping the Cygwin maintainer (although I think that may involve offering to become the Cygwin maintainer......)

xavierleroy:
With my CompCert hat on, I would like to use OCaml under Cygwin64.  How could we make progress with the new release of FlexDLL and the updated Cygwin64 package?

dra27:
I'm just testing 0.39 on my fork - assuming that's fine, I'll do the release later today.

dra27:
It's released, and I've proposed [adopting the Cygwin package](https://cygwin.com/pipermail/cygwin-apps/2020-November/040734.html) to push the release downstream as well.

PR #90, opened 2020-09-21
Fix flexdll_dlopen when passed NULL file

dra27:
On native Windows (i.e. not Cygwin), flexdll_dlopen is a stub function for converting an ANSI filename to Unicode and then calling
flexdll_wdlopen. However, it's permitted to pass NULL in order to get a handle to the running process, but this isn't implemented in the stub, which fails.

As far as I can tell, this means ocamlnat has been broken since OCaml 4.06!


alainfrisch (approved):
LGTM

dra27:
Ta!

PR #93, opened 2021-09-19
Add "incbin" functionality

nojb:
This PR adds an "incbin" (see https://github.com/graphitemaster/incbin) feature to `flexdll`, allowing to include binary blobs in the final binary. The motivation is https://github.com/ocaml/ocaml/pull/10159 and https://github.com/ocaml/ocaml/pull/10654, namely including debug information for bytecode binaries built with `--output-complete-exe` with MSVC, but the functionality makes sense independently.

Concretely, a new flag `-incbin <sym> <file>` is added which has the effect of embedding `<file>` inside the final binary and defining two symbols: `<sym>_data` and `<sym>_size`: `<sym>_data` points to the beginning of the binary blob, and `<sym>_size` is a 32-bit int containing the size of the binary blob.

I'm still testing it, but feedback would be welcome!

### DIFF
--- a/cmdline.ml
+++ b/cmdline.ml
@@ -37,6 +37,7 @@ let output_file = ref ""
 let exe_mode : [`DLL | `EXE | `MAINDLL] ref = ref `DLL
 let extra_args = ref []
 let mode : [`NORMAL | `DUMP | `PATCH] ref = ref `NORMAL
+let incbins = ref []
 let defexports = ref []
 let noentry = ref false
 let use_cygpath = ref true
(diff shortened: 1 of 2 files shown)


dra27:
It's not a strong enough objection to say "this shouldn't be in flexlink", but it does a feel little as though a _compiler_ function is being added to a _linker_ largely because the linker has the `Coff` module 🙂

I'm vaguely wondering if either `#embed` (C23, so it might be prudent not to hold our breath!) or a solution based on resource files might be better for this in the long run? I don't have any performance evidence, but I'm wondering if there's a reason that PDB is _always_ in a separate file i.e. while in ELF we can clearly have the debugging information present with no issue, I'm left wondering why PE goes to great lengths to _reference_ external debugging data, rather than ever allowing it to be embedded.

nojb:
You are right; I'm also not convinced of the approach (I closed https://github.com/ocaml/ocaml/pull/10654). Let's close this for now.

PR #110, opened 2022-11-17
Cosmetic updates

dra27:
- Eliminate spurious error messages when `ocamlopt` isn't available in the environment
- Missing `.gitignore` patterns for release tarballs and zips
- Tidy up `LICENSE` file
- Add additional fields to the version information block in `flexlink.exe`
- Add a `-vnum` and `-version` command line argument (finally)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,8 @@
 *.cmo
 *.cmi
 *.dll
+*.tar.gz
+*.zip
 flexlink.exe
 flexlink.opt
 test/dump.exe
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-The package FlexDLL is released under the terms of a zlib/libpng License.
+The package FlexDLL is released under the terms of the zlib/libpng License.
 
 Copyright (c) 2007, 2008, 2009
 Institut National de Recherche en Informatique et en Automatique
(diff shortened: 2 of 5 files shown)


nojb:
Hello @dra27! I just saw this PR fly past and thought I would take advantage of it to pose an unrealted question. We maintain the following diff for one of our clients (who builds OCaml from source, with the "boostrapped" FlexDLL):
```diff
diff --git a/mlfi/ocaml/flexdll/Makefile b/mlfi/ocaml/flexdll/Makefile
--- a/mlfi/ocaml/flexdll/Makefile
+++ b/mlfi/ocaml/flexdll/Makefile
@@ -161,7 +161,7 @@ Compat.ml: COMPILER-$(COMPAT_VERSION) $(addsuffix .ml, $(COMPAT_MODULES))
 flexlink.exe: $(OBJS) $(RES)
        @echo Building flexlink.exe with TOOLCHAIN=$(TOOLCHAIN) for OCaml $(OCAML_VERSION)
        rm -f flexlink.exe
-       $(RES_PREFIX) $(OCAMLOPT) -o flexlink.exe $(LINKFLAGS) $(OBJS)
+       FLEXLINKFLAGS="" $(RES_PREFIX) $(OCAMLOPT) -o flexlink.exe $(LINKFLAGS) $(OBJS)

 version.res: version.rc
        $(RES_PREFIX) rc version.rc
```
The point is that they set `FLEXLINKFLAGS` to pass some flags to `flexlink` when building the compiler (if memory serves, it is `-link <resource file>` to add icons and copyright information to the built objects), but this variable should be _unset_ when building `flexlink` itself. As far as I can see there is no way to do this from "outside" the build system. Does it make sense? Do you see how to achieve this, or a better way to go about it?

PR #112, opened 2023-02-20
Put error global variables into thread-local storage

shym:
Global variable `error_buffer` is used to store a string that is returned to callers, so there is a race condition if dynamic linking is invoked from 2 OCaml domains in parallel
Since the error message must be returned, a mutex cannot be used to prevent the race condition
GNU libc uses that same solution: keep the last error in thread-local storage; so support for calling dlerror from a different thread than the one calling dlopen is not to be expected

The problem with parallel accesses was discovered investigating segfaults in ocaml-multicore/multicoretests#290.
With this patch, these tests do not segfault.
I see the following 4 dynlink tests from the ocaml test suite failing, but they seem to fail whether or not this patch is applied and the error message seems not related.
```
> run_win32.c:365: CreateProcess failed: The system cannot find the file specified.
[…]
List of failed tests:
    tests/lib-dynlink-csharp /'main.ml' with 1.1.4.1.1.1.1 (script)
    tests/lib-dynlink-csharp /'main.ml' with 1.1.3.1.1.1 (script)
    tests/lib-dynlink-csharp /'main.ml' with 1.1.2.1.1.1.1 (script)
    tests/lib-dynlink-csharp /'main.ml' with 1.1.1.1.1.1 (script)
```

### DIFF
--- a/flexdll.c
+++ b/flexdll.c
@@ -47,12 +47,98 @@ typedef struct dlunit {
 } dlunit;
 typedef void *resolver(void*, const char*);
 
+/* Error reporting */
+/* The latest error must be kept in some variable so that flexdll_dlerror can
+ * report it but this causes data races (and possible segmentation faults) in
+ * multithreaded programs if that variable is global. So this must use
+ * thread-local storage instead.
+ *
+ * To ensure compatibility, the implementation does not require compiler support
+ * for thread-local storage (__thread) and instead relies on functions
+ * TlsGetValue, TlsSetValue, etc.
+ *
+ * This implementation is structured around the function get_tls_error that
+ * returns a structure containing, for the current thread, the code (and a
+ * buffer for the corresponding message) of the last flexdll error. It accepts
+ * an argument to choose whether the structure should be reset:
+ *
+ * - TLS_ERROR_RESET will reset what is stored in the structure, so this is
+ *   intended for initialisation entry points (flexdll_dlopen, flexdll_relocate)
+ * - TLS_ERROR_NOP will keep the current content of the structure, for all other
+ *   entry points (flexdll_dlerror, ll_dlerror)
+ *
+ * The other exported entrypoints do not need to access the error storage.
+ */
+
 typedef struct error_s {
   int code;
   char message[256];
 } err_t;
 
-static err_t err;
+#define TLS_ERROR_NOP 0
+#define TLS_ERROR_RESET 1
+
+#define TLS_ACCESS_ERRMSG "error accessing thread-local storage"
+
+static err_t *get_tls_error(int op) {
+  static volatile DWORD error_idx = TLS_OUT_OF_INDEXES;
+  DWORD new_idx, last_error;
+  err_t *error;
+
+  /* We store the last system error to restore it on leaving, so that
+   * get_tls_error is transparent with regards to GetLastError */
+  last_error = GetLastError();
+
+  if(error_idx == TLS_OUT_OF_INDEXES) {
+    new_idx = TlsAlloc();
+    if(new_idx == TLS_OUT_OF_INDEXES) {
+      /* If we cannot allocate the structure required to report errors... */
+      /* Maybe we should set the last error to some standard value
+       * that correspond to such cases instead of resetting it? */
+      SetLastError(last_error);
+      return NULL;
+    }
+    /* According to documentation DWORD and LONG take both 32 bits so
+     * this uses InterlockedCompareExchange to store a DWORD */
+    if (InterlockedCompareExchange((LONG*)&error_idx, (LONG)new_idx, (LONG)TLS_OUT_OF_INDEXES) != (LONG)TLS_OUT_OF_INDEXES) {
+      if(!TlsFree(new_idx)) {
+        SetLastError(last_error);
+        return NULL;
+      }
+    }
+  }
+
+  error = TlsGetValue(error_idx);
+  if(error == NULL) {
+    error = malloc(sizeof(err_t));
+    if(error == NULL) {
+      SetLastError(last_error);
+      return NULL;
+    }
+    if(!TlsSetValue(error_idx, error)) {
+      free(error);
+      SetLastError(last_error);
+      return NULL;
+    }
+    error->code = 0;
+    error->message[0] = 0;
+  }
+
+  SetLastError(last_error);
+
+  switch(op) {
+  case TLS_ERROR_NOP:
+    break;
+  case TLS_ERROR_RESET:
+    error->code = 0;
+    error->message[0] = 0;
+    break;
+  default:
+    return NULL;
+  }
+
+  return error;
+}
 
 /* Emulate a low-level dlopen-like interface */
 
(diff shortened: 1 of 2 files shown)


dra27:
(closed and re-opened to recompute the merge commit now that #115 is merged, so CI should have something useful to say!)

flexdll.c
```diff
@@ -47,8 +47,8 @@ typedef struct dlunit {
 } dlunit;
 typedef void *resolver(void*, const char*);
 
-static int error = 0;
-static char error_buffer[256];
+static __thread int error = 0;
+static __thread char error_buffer[256];
```
dra27, line 51: I think this should work:
```suggestion
#ifdef _MSC_VER
#define __thread __declspec(thread)
#endif
static __thread int error = 0;
static __thread char error_buffer[256];
```

dra27:
Hmm, this is looking tedious... it looks like we're pulling in a runtime library somewhere from the error. I'm guessing that `/usr/i686-w64-mingw32/sys-root/mingw/bin` (and the x64 equivalent) need adding to PATH for this to work, although we should nail down exactly which DLL it's trying to pull in. That's a bit too heavy for this - we might instead need to hand-roll the native Windows version using `TlsAlloc` et al.

shym:
I updated the PR with an implementation using `Tls*` functions. This turned out a bit more involved than what I first thought, since `TlsGetValue` modifies the result of `GetLastError` which we want to preserve. I hope the comments are enough to clarify the implementation choices.

Reproducing locally the test that failed, I get a pop-up saying `libwinpthreads-1.dll` is missing, maybe too big a dependency for that single feature.

nojb:
Hello @shym: just a heads-up that I am planning to read this PR soon. Thank you for your patience!

flexdll.c
```diff
@@ -47,8 +47,99 @@ typedef struct dlunit {
...
+    new_idx = TlsAlloc();
+    if(new_idx == TLS_OUT_OF_INDEXES)
+      /* If we cannot allocate the way to report errors... */
+      return NULL;
+    /* According to documentation DWORD and LONG take both 32 bits so
+     * this uses InterlockedCompareExchange to store a DWORD */
+    if (InterlockedCompareExchange((LONG*)&error_idx, (LONG)new_idx, (LONG)TLS_OUT_OF_INDEXES) != (LONG)TLS_OUT_OF_INDEXES) {
+      if(!TlsFree(new_idx))
+        return NULL;
+      goto retry;
```
nojb, line 111: If we reach this `goto`, can it happen that the condition at the toop of the loop `error_idx == TLS_OUT_OF_INDEXES` holds? If not, can we get rid of the `goto` altogether?

nojb (dismissed):
Thank you for this patch! The code is extremely clear and a pleasure to review :)

LGTM (modulo a small question)

flexdll.c
shym: Good point!
I didn't find anything in the documentation that would invite to retest and local tests seem to work. I had followed this [example](https://github.com/dra27/flexdll/commit/7bbaf6649922d011df7efc24f9c9427bb4bc19c9?diff=unified#diff-6da5df65a6271e4d53ced90f2bb6d89edcdf56ddf1c1c13d79b007ed572fbfdcR400) but that commit was explicitly and intentionally over-defensive, so I also think we can simplify.

shym:
Thanks you for your kind and thorough review!
I've updated the branch removing the `goto`.
Looking at another PR, I should add an entry to `CHANGES`, maybe?

flexdll.c
```diff
@@ -357,9 +451,13 @@ static void *find_symbol_global(void *data, const char *name) {
 }
 
 int flexdll_relocate(void *tbl) {
+  err_t * err;
+  err = get_tls_error(TLS_ERROR_RESET_LAST);
+  if(err == NULL) return 0;
```
nojb, line 456: Thinking more about this, shouldn't we reset `err->code = 0` here? Otherwise, the check below in line 460 will fail if this function is called after another function that has set `err->code`.

Going one step further, perhaps when in `TLS_ERROR_RESET_LAST` mode, we should always set `err->code = 0`. Or is there a case where we want to reset one of the error codes, but not the other?

shym: Very good point, thank you very much!

I think I ended up with that code because it was not explicitly reset in the original code. That was arguably correct because `flexdll_relocate` is called from two places (if I didn't miss any other):
- from `flexdll_dlopen` where `code` has already been reset,
- from `flexdll_init` where `code` has been set to its initial value 0.
This made me realize that I had forgotten to initialize the values when they are malloc-ed!

So I’ve rewritten the code so that:
- on `TLS_ERROR_RESET`, both `code` and `last_error` are reset (so no `_LAST`),
- the explicit reset of `code` near the call to `get_tls_error(TLS_ERROR_RESET)` are removed,
- the structure is initialized right after `malloc`, just to make sure; the structure _should_ be malloc-ed on a call to one of the initialisation entrypoints, in which case it will be initialised again just a few lines later, but that will ensure that a buggy program calling `dlerror` without a previous call to `dlopen` will get a reliable reasonable behaviour.

nojb (commented):
Sorry for the back-and-forth, but it turns out that there is a function `SetLastError` https://learn.microsoft.com/en-us/windows/win32/api/errhandlingapi/nf-errhandlingapi-setlasterror
Couldn't we use that instead to restore the result of the call to `GetLastError` after calling the `Tls*` functions? It should simplify the code (no need for the `last_error` field).

shym:
Very good idea indeed!
I noticed that the documentation for `SetLastError` explicit states that the last error is stored in TLS and that values with bit 29 set are reserved for user errors. But I didn’t find a trick to reuse that to fully skip using TLS explicitly ourselves, especially since POSIX’s `dlerror` must report the last error of a dl function, so other functions must not interfere with the result it will report.
So I just updated the patch removing the `last_error` field and merging all the non-resetting behaviours. It’s a lot simpler to read.

flexdll.c
```diff
@@ -462,11 +565,15 @@ void *flexdll_dlsym(void *u, const char *name) {
 }
 
 char *flexdll_dlerror() {
-  switch (error) {
+  err_t * err;
+  err = get_tls_error(TLS_ERROR_NOP);
+  if(err == NULL) return "error in accessing thread-local storage";
```
nojb, line 570:
```suggestion
  if(err == NULL) return "error accessing thread-local storage";
```

flexdll.c
```diff
@@ -104,16 +192,21 @@ static void *ll_dlsym(void *handle, char *name) {
 
 static char *ll_dlerror(void)
 {
-  DWORD msglen =
+  DWORD msglen;
+  err_t * err;
+  err = get_tls_error(TLS_ERROR_NOP);
+  if(err == NULL) return "error in accessing thread-local storage";
```
nojb, line 198:
```suggestion
  if(err == NULL) return "error accessing thread-local storage";
```

nojb (approved):
Looks good to me, thanks!

shym:
Just changed the error message, and added due credit! :smile:

nojb:
Thanks, merged! (I took the liberty of squashing the commits into a single commit; this makes it easier to revert, cherry-pick, etc.)

PR #117, opened 2023-04-15
Coff archives: handle newline-separated long names

### LINKED ISSUE #101: Implementing a less memory intensive read function
When building the ocaml bibindings for Z3, it attempts to allocate ~10GiB before an oom error. This exceedingly high memory usage occurs while reading the import library (libz3.dll.a), and before the actual linker is called.

I don't know if there's a more efficient way to store these strings,  or if it's possible to implement it such as to save results to disk and continue reading if memory usage is becoming too high. It seems very improbable that  33 GiB of strings are needed to make ~500KiB to 1.3MiB libraries.

The most pertinent information from the ctf is likely:
```
1391 samples of 10.4 GB allocations

 5.9 GB (57.0%) at Stdlib__Bytes.sub (bytes.ml:68:12-22, 68:12-22)
      including:
     2.9 GB via Coff.Lib.read_lib.read_member (97% of this)

 4.4 GB (42.4%) at Stdlib__Buffer.resize (buffer.ml:87:19-40)
      including:
     4.1 GB via Coff.Lib.read_lib.obj (56% of this)
     4.1 GB via Stdlib__Printf.ksprintf.k' (93% of this)
     4.1 GB via CamlinternalFormat.strput_acc (93% of this)
     4.1 GB via Stdlib__Buffer.add_string (93% of this)
```

The CTF is here, if you'd like to view it, though. 
https://gist.github.com/EmmaJaneBonestell/98fda00c93a8d09ea09bada08972761d 

Edit: I used a GCP server, and it took about 33 GiB of RAM and a total of 46 GiB of allocations to build. 

Edit 2: UCRT had nothing to do with it, but rather, it seems to have the excessive memory usage with any of the MSYS2 environments' built dlls. If the dll is built with Cygwin's MinGW toolchain, rather than MSYS2's MINGW toolchain, it doesn't have a problem. E.g. libz3.dll.a built with Cygwin's MinGW, vs MSYS2's, won't consume excessive RAM.

nojb:
Static archives (`.a`) generated by GNU toolchains use newline (`\n`) as separator in the "long names" header, instead of `\000` (used in MS' toolchain). This header is used whenever object names take more than 16 bytes. The absence of `\000`'s triggered a quadratic behaviour when reading object names.

To fix this, we postprocess the "long name" string table to replace `\n` (and/or the optional trailing `/`) by `\000`. Similar logic is present in `binutils`:

https://github.com/bminor/binutils-gdb/blob/672008e6dcacddd76a5551aec9a30cd31c102d72/bfd/archive.c#L1323-L1324

(`ARFMAG[1]` is `\n` above.)

Fixes #101

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -2,6 +2,7 @@ Next version
 - GPR#108: Add -lgcc_s to Cygwin's link libraries, upstreaming a patch from the
   Cygwin flexdll package (David Allsopp)
 - GPR#114: Support for /alternatename directive. Fixes GPR#113 (Jonah Beckford)
+- GPR#117: Fix handling of object names of length > 16 in archive files. Fixes GPR#110 (Nicolás Ojeda Bär)
 
 Version 0.42
 - GPR#106: Support -l: syntax, to allow static linking of specific libraries (David Allsopp)
--- a/coff.ml
+++ b/coff.ml
@@ -950,7 +950,17 @@ module Lib = struct
       let name = strz (Bytes.sub buf 0 16) 0 ' ' in
       begin match name with
         | "/" | "" -> ()
-        | "//" -> strtbl := read ic (pos_in ic) size
+        | "//" ->
+            let tbl = read ic (pos_in ic) size in
+            (* GNU-produced archive files are supposed to be printable if the
+               archive members contain text, so the entries in the list are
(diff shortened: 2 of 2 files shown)


nojb:
In the absence of @dra27 who is out sick, I am merging this bugfix that is rather low-risk.

Issue #41, opened 2017-09-19
Binary release?


damiendoligez:
Where can I find a binary installer of flexdll 0.36?

alainfrisch:
Ok, I've uploaded my first attempt at binary releases on https://github.com/alainfrisch/flexdll/releases/tag/0.36 (both the .zip and the installer).

They have been built with the 32-bit mingw port of OCaml 4.06.  This means that flexlink.exe should be Unicode aware and that it should run about anywhere (on 32-bit machines, and on machines without any MSVC redistributables).

The msvc support files (flexdll_msvc.obj and co) have been built with VS2015.  My guess is that they would not work with older or newer versions of Visual Studio,  but I might be wrong.  @dra27?

alainfrisch:
@damiendoligez Did you need to get the binary distrib for testing purposes, or to update README.win32.adoc?

nojb:
My understanding is that objects built with VS 2015 will work with newer versions, but not older.

damiendoligez:
I needed it for updating the CI machines, but I ended up recompiling flexdll myself, which is a good thing because the CI machines are on VS 2013 for the moment.

dra27:
Just for "ease", I'd keep the binary release using VS < 2015

alainfrisch:
Anyway, I screwed up the release, the msvc64 objects have been compiled with the 32-bit cl (I used MSVC_DETECT=0, shame on me), so I'll need to do it again.

What do you mean by "ease"?  This will certainly be easier for users with VS < 2015, and one can guess that most OCaml users are in this case (because that's what previous binary releases of flexdll supported).  But shouldn't we give an incentive to switch to newer versions (with the hope to be able to get rid of supporting older versions of VS at some point)?

I thought also on the following: one could consider VS < 2015 and VS >= 2015 as two different toolchains for flexdll, and ship *.obj files for both in the binary release.

nojb:
In the short term, I think that shipping both versions would be best for "ease".

But longer term, wouldn't it make sense to compile the support code as part of the OCaml compilation process (using the same toolchain as the rest of the compilation process) and install the object files in the OCaml library directory, where flexlink could look for them ?

dra27:
@alainfrisch wrote:
> Anyway, I screwed up the release, the msvc64 objects have been compiled with the 32-bit cl (I used MSVC_DETECT=0, shame on me), so I'll need to do it again.

I did propose having a configure script :wink:

[...]

PR #124, opened 2023-11-23
Update AppVeyor scripts

MisterDA:
- VS 2022
- OCaml 5.1
- FlexDLL 0.43

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -10,6 +10,7 @@ environment:
     CYG_CACHE: C:/cygwin64/var/cache/setup
     OCAML_PORT: msvc64
     ARTEFACTS: no
+    FLEXDLL_VERSION: 0.43
   matrix:
     - OCAMLBRANCH: 3.11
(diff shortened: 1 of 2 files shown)


appveyor.yml
```diff
@@ -1,7 +1,7 @@
 platform:
   - x64
 
-image: Visual Studio 2015
+image: Visual Studio 2022
```
dra27, line 4: See the `ARTEFACTS: yes` case below... this will almost certainly need to use an old image to maintain the binary releases as they are presently done (see #41 and #46).

Given that VS 2015 is still supported, I'm not sure that the switch to VS 2022 is worth it?

dra27 (commented):
Not sure about the VS 2022 part, but the other two bits are definitely good, thanks!

appveyor.yml
MisterDA: You're likely right, I'm reverting that part.

dra27:
Thanks!

PR #125, opened 2023-12-11
Use Make Automatic Variables, update msvs-detect

MisterDA:
Hacking in passing, I decided to use Make Automatic Variables to avoid redundancies with file names, and update msvc-detect.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -223,7 +223,7 @@ distclean: clean
 
 clean:
 	rm -f *.obj *.o *.lib *.a *.exe *.opt *.cmx *.dll *.exp *.cmi *.cmo *~ version.res version.ml COMPILER-* Compat.ml
-	cd test && $(MAKE) clean
+	$(MAKE) -C test clean
 
 
 ## Packaging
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -91,6 +93,11 @@ MSVCC64 = cl.exe $(MSVC_FLAGS)
 endif
 endif
 
+ifeq ($(shell expr "$(CI)" : '^[Tt]rue$$'),4)
```
dra27, line 96: I think this avoids the shell call (which matters a lot to me... cf. #119)
```suggestion
ifneq "$(filter True true,$(CI))" ""
```

dra27 (commented):
In-passing comment on one of the Makefile changes!

I've looked at some of the earlier commits of this already - I like the direction it's going, but could you split the first 3 commits to a separate PR because I think they should all pass immediately and it will get the noise out of the way of looking at the C compiler warnings part?

MisterDA:
I've removed the compiler warnings commits from this PR.

dra27 (approved):
Lovely, ta!

PR #127, opened 2023-12-18
Fix failure while parsing '-' in directive

rzabarazesh:
This fixes https://github.com/ocaml/flexdll/issues/126


dra27 (approved):
LGTM, thanks! The COFF spec is vague on the exact format (bordering on incorrect, given that it says options _should_ begin with `-` yet the MS C compiler embeds options beginning `/`!). I can see that this capability was added to GNU ld within the last 18 months, so I guess we're just starting to see it trickle through in actual objects.

PR #129, opened 2024-01-17
Add and fix C compiler warnings

MisterDA:
I noticed that the `CPPFLAGS` and `CFLAGS` were not being passed when building Flexdll's C files, but they introduce problems when building OCaml 4.06 and 4.07 (explanation from David Allsopp):

- The underlying OCaml compiler is msvc64, unless `OCAML_PORT` is set in AppVeyor matrix (so OCaml in the 4.06 and 4.07 tests is msvc64);
- https://github.com/ocaml/ocaml/pull/1114 (4.06.0) introduced `CFLAGS` and `CPPFLAGS` in the OCaml's installed Makefile.config;
- FlexDLL's Makefile includes that `$(ocamlopt -where)/Makefile.config` so ends up getting OCaml's msvc64 `CFLAGS` and `CPPFLAGS` values;
- https://github.com/ocaml/ocaml/pull/1840 (4.08.0) renamed them to `OC_CFLAGS` and `OC_CPPFLAGS` which removes the problem.

I've instead chosen to add `-Wall` to enable warnings when compiling C files. To make sure new problems are not introduced, the warnings are turned into errors if and only if the `CI` environment variable matches `[Tt]rue`.

I've then tentatively fixed the warnings raised by GCC and MSVC.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -97,6 +97,12 @@ MSVCC64 = cl.exe $(MSVC_FLAGS)
 endif
 endif
 
+FLEXDLL_WARN_ERROR ?=
+ifeq ($(FLEXDLL_WARN_ERROR),true)
+GCC_FLAGS += -Werror -fdiagnostics-color=always
+MSVC_FLAGS += /WX
+endif
+
 show_root:
 ifeq ($(MSVCC_ROOT),)
 	@echo "$(MSVS_PATH)"
--- a/Makefile
+++ b/Makefile
@@ -26,14 +26,16 @@ COMPAT_VERSION:=$(subst ., ,$(OCAML_VERSION))
 COMPAT_VERSION:=$(subst $(SPACE),,$(firstword $(COMPAT_VERSION))$(foreach i,$(wordlist 2,$(words $(COMPAT_VERSION)),$(COMPAT_VERSION)),$(if $(filter 0 1 2 3 4 5 6 7 8 9,$(i)),0,)$(i)))
 endif
 
+GCC_FLAGS = -Wall
+
 MINGW_PREFIX = i686-w64-mingw32-
-MINCC = $(MINGW_PREFIX)gcc
+MINCC = $(MINGW_PREFIX)gcc $(GCC_FLAGS)
 
 MINGW64_PREFIX = x86_64-w64-mingw32-
-MIN64CC = $(MINGW64_PREFIX)gcc
+MIN64CC = $(MINGW64_PREFIX)gcc $(GCC_FLAGS)
 
 CYGWIN64_PREFIX = x86_64-pc-cygwin-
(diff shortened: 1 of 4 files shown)


flexdll.c
```diff
@@ -434,6 +433,7 @@ static void unlink_unit(dlunit *unit) {
 static void *find_symbol_global(void *data, const char *name) {
   void *sym;
   dlunit *unit;
+  (void)data;
```
jmid, line 436: This puzzled me a bit. I suppose this is needed to avoid an unused parameter warning?

jmid (approved):
LGTM! It is already showing its value by having identified, e.g., unused local variables.

flexdll.c
MisterDA: Yes, exactly.

Makefile
```diff
@@ -83,16 +85,21 @@ MSVC_PREFIX=LIB="$(MSVC_LIB)" INCLUDE="$(MSVC_INCLUDE)"
...
 MSVC64_PREFIX=PATH="$(MSVS64_PATH)$(PATH)" LIB="$(MSVS64_LIB)$(LIB)" INCLUDE="$(MSVS64_INC)$(INCLUDE)"
 
-MSVCC = cl.exe $(MSVC_FLAGS)
-MSVCC64 = cl.exe $(MSVC_FLAGS)
+MSVCC = cl.exe $(MSVC_CFLAGS)
+MSVCC64 = cl.exe $(MSVC_CFLAGS)
+endif
 endif
+
+ifneq "$(filter True true,$(CI))" ""
```
dra27, line 100: I didn’t think of this previously - unlike where we were using `GITHUB_ACTIONS` in ocaml/ocaml#12842, but we need to be sure that this can only be triggered by a workflow/CI run _here_ - at the moment, I think this will enable warnings as errors on all GitHub Actions runs building the compiler?

I think therefore should be a variable which is set by the AppVeyor script e.g. adding `FLEXDLL_WARN_ERROR: true` to the `global` section of `environment` in appveyor.yml and then having:

```suggestion
FLEXDLL_WARN_ERROR ?=
ifneq "$(filter True true,$(FLEXDLL_WARN_ERROR))" ""
```

flexdll.c
dra27: It had puzzled me in the draft review I started a while back too! Please could there be a comment, therefore?

dra27 (commented):
Apologies, I started a review of this several months ago, and never finished it 😕

Two very minor tweaks and this'll be good to go, thank you!

flexdll.c
MisterDA: Added a comment to that effect.

dra27 (approved):
Lovely, thank you very much!

Issue #120, opened 2023-06-08
Fix parallel access to global symbol structures


dra27:
The list of loaded units is defined globally in `flexdll.c`:
https://github.com/ocaml/flexdll/blob/cc0525e99e7109c8d3206085759185bd733ac49f/flexdll.c#L417

and is accessed in parallel both by `flexdll_*dlopen` and `flexdll_dlsym` without using a lock. Adding locks around `flexdll_*dlopen` is straightforward, but symbol lookup operates a most-recently-used queue on the global, and the performance impact either of removing that or, worse, of putting a lock around `flexdll_dlsym` is less clear.

In the meantime, reverting ocaml/ocaml#11607 provides an easy test-case for demonstrating the failure (the test usually fails within a few runs)

jmid:
This was fixed in #136 and included in the 0.44 release.
The issue can now be closed.

dra27:
Indeed!

PR #136, opened 2024-04-16
Fix parallel access crashes and misbehavior

jmid:
Parallel usage is memory unsafe (read: may crash) as documented in #120, ocaml/ocaml#11607, and ocaml/ocaml#13046.

This PR goes for the simplest possible fix: adding a single global lock by dusting off the first commit of https://github.com/dra27/flexdll/tree/sledgehammer and suitable rebasing, renaming, and error handling.
Author credit thus goes to @dra27 - any errors are mine.

For the error handling, I've tried to make it fit with @shym's TLS-based error handling from #112.
I'm unsure how to test these error code paths though without explicitly mocking with the source code to create an invalid lock handle.

With the fix
- the ocaml/ocaml testsuite passes incl. the disabled `tests/lib-dynlink-domains` test from ocaml/ocaml#11607
- the reproducer from ocaml/ocaml#13046 also passes
- the `Dynlink` stress test from `multicoretests` passes

(these have been tested under MinGW in a Cygwin-shell)

### DIFF
--- a/flexdll.c
+++ b/flexdll.c
@@ -501,6 +501,10 @@ void *flexdll_wdlopen(const wchar_t *file, int mode) {
 again:
   if (units_mutex == INVALID_HANDLE_VALUE) {
     HANDLE hMutex = CreateMutex(NULL, TRUE, NULL);
+    if (hMutex == NULL) {
+      if (!err->code) err->code = 1;
+      return NULL;
+    }
     if (InterlockedCompareExchangePointer(&units_mutex, hMutex, INVALID_HANDLE_VALUE) != INVALID_HANDLE_VALUE) {
       CloseHandle(hMutex);
       goto again;
--- a/flexdll.c
+++ b/flexdll.c
@@ -509,11 +509,9 @@ again:
       CloseHandle(hMutex);
       goto again;
     }
-  } else {
-    if (WaitForSingleObject(units_mutex, INFINITE) == WAIT_FAILED) {
+  } else if (WaitForSingleObject(units_mutex, INFINITE) == WAIT_FAILED) {
       if (!err->code) err->code = 1;
       return NULL;
-    }
   }
 
   handle = ll_dlopen(file, exec);
--- a/flexdll.c
+++ b/flexdll.c
@@ -582,7 +582,7 @@ void *flexdll_dlsym(void *u, const char *name) {
   void *res;
   err_t * err;
   err = get_tls_error(TLS_ERROR_NOP);
-  if(err == NULL) return NULL;
+  if (err == NULL) return NULL;
 
   if (WaitForSingleObject(units_mutex, INFINITE) == WAIT_FAILED) {
     if (!err->code) err->code = 1;
--- a/flexdll.c
+++ b/flexdll.c
@@ -67,7 +67,7 @@ static volatile HANDLE units_mutex = INVALID_HANDLE_VALUE;
  * - TLS_ERROR_RESET will reset what is stored in the structure, so this is
  *   intended for initialisation entry points (flexdll_dlopen, flexdll_relocate)
  * - TLS_ERROR_NOP will keep the current content of the structure, for all other
- *   entry points (flexdll_dlerror, ll_dlerror)
+ *   entry points (flexdll_dlerror, ll_dlerror, flexdll_dlsym)
  *
  * The other exported entrypoints do not need to access the error storage.
  */
--- a/flexdll.c
+++ b/flexdll.c
@@ -47,7 +47,7 @@ typedef struct dlunit {
 } dlunit;
 typedef void *resolver(void*, const char*);
 
-static volatile HANDLE what_was_I_thinking = INVALID_HANDLE_VALUE;
+static volatile HANDLE units_mutex = INVALID_HANDLE_VALUE;
 
 /* Error reporting */
 /* The latest error must be kept in some variable so that flexdll_dlerror can
@@ -499,14 +499,14 @@ void *flexdll_wdlopen(const wchar_t *file, int mode) {
 #endif /* CYGWIN */
 
 again:
-  if (what_was_I_thinking == INVALID_HANDLE_VALUE) {
+  if (units_mutex == INVALID_HANDLE_VALUE) {
     HANDLE hMutex = CreateMutex(NULL, TRUE, NULL);
-    if (InterlockedCompareExchangePointer(&what_was_I_thinking, hMutex, INVALID_HANDLE_VALUE) != INVALID_HANDLE_VALUE) {
+    if (InterlockedCompareExchangePointer(&units_mutex, hMutex, INVALID_HANDLE_VALUE) != INVALID_HANDLE_VALUE) {
       CloseHandle(hMutex);
       goto again;
     }
   } else {
-    if (WaitForSingleObject(what_was_I_thinking, INFINITE) == WAIT_FAILED) {
+    if (WaitForSingleObject(units_mutex, INFINITE) == WAIT_FAILED) {
       /* XXX Some kind of error here?? */
       /* error = ?? */
       return NULL;
@@ -514,7 +514,7 @@ again:
   }
 
   handle = ll_dlopen(file, exec);
-  if (!handle) { if (!err->code) err->code = 1; ReleaseMutex(what_was_I_thinking); return NULL; }
+  if (!handle) { if (!err->code) err->code = 1; ReleaseMutex(units_mutex); return NULL; }
 
   unit = units;
   while ((NULL != unit) && (unit->handle != handle)) unit = unit->next;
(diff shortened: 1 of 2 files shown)


flexdll.c
```diff
@@ -496,8 +498,22 @@ void *flexdll_wdlopen(const wchar_t *file, int mode) {
...
+    if (InterlockedCompareExchangePointer(&units_mutex, hMutex, INVALID_HANDLE_VALUE) != INVALID_HANDLE_VALUE) {
+      CloseHandle(hMutex);
+      goto again;
+    }
+  } else {
+    if (WaitForSingleObject(units_mutex, INFINITE) == WAIT_FAILED) {
+      if (!err->code) err->code = 1;
+      return NULL;
+    }
+  }
```
MisterDA, line 513: I find the `goto again` a bit convoluted, isn't a simple while loop equivalent?

```suggestion
  while (units_mutex == INVALID_HANDLE_VALUE) {
    HANDLE hMutex = CreateMutex(NULL, TRUE, NULL);
    if (hMutex == NULL) /* do something */;
    if (InterlockedCompareExchangePointer(&units_mutex, hMutex, INVALID_HANDLE_VALUE) != INVALID_HANDLE_VALUE)
      CloseHandle(hMutex);
    else break; /* optional */
  }
  
  if (WaitForSingleObject(units_mutex, INFINITE) == WAIT_FAILED) {
    if (!err->code) err->code = 1;
      return NULL;
    }
  }
```

flexdll.c
```diff
@@ -47,6 +47,8 @@ typedef struct dlunit {
 } dlunit;
 typedef void *resolver(void*, const char*);
 
+static volatile HANDLE units_mutex = INVALID_HANDLE_VALUE;
```
MisterDA, line 50: Is the `volatile` really needed?

flexdll.c
```diff
@@ -496,8 +498,22 @@ void *flexdll_wdlopen(const wchar_t *file, int mode) {
...
 
+again:
+  if (units_mutex == INVALID_HANDLE_VALUE) {
+    HANDLE hMutex = CreateMutex(NULL, TRUE, NULL);
+    if (InterlockedCompareExchangePointer(&units_mutex, hMutex, INVALID_HANDLE_VALUE) != INVALID_HANDLE_VALUE) {
+      CloseHandle(hMutex);
+      goto again;
+    }
+  } else {
+    if (WaitForSingleObject(units_mutex, INFINITE) == WAIT_FAILED) {
```
MisterDA, line 509:
```suggestion
    if (WaitForSingleObject(units_mutex, INFINITE) != WAIT_OBJECT_0) {
```
This would cover the improbable (impossible?) case of the mutex having been abandoned.
https://learn.microsoft.com/en-us/windows/win32/sync/using-mutex-objects

flexdll.c
```diff
@@ -561,9 +579,20 @@ void flexdll_dlclose(void *u) {
...
 void *flexdll_dlsym(void *u, const char *name) {
-  if (u == &main_unit) return find_symbol_global(NULL,name);
-  else if (NULL == u) return find_symbol(&static_symtable,name);
-  else return find_symbol(((dlunit*)u)->symtbl,name);
+  void *res;
+  err_t * err;
+  err = get_tls_error(TLS_ERROR_NOP);
+  if(err == NULL) return NULL;
+
+  if (WaitForSingleObject(units_mutex, INFINITE) == WAIT_FAILED) {
```
MisterDA, line 587:
```suggestion
  if (WaitForSingleObject(units_mutex, INFINITE) == WAIT_OBJECT_0) {
```

flexdll.c
```diff
@@ -561,9 +579,20 @@ void flexdll_dlclose(void *u) {
 
 
 void *flexdll_dlsym(void *u, const char *name) {
-  if (u == &main_unit) return find_symbol_global(NULL,name);
-  else if (NULL == u) return find_symbol(&static_symtable,name);
-  else return find_symbol(((dlunit*)u)->symtbl,name);
+  void *res;
+  err_t * err;
+  err = get_tls_error(TLS_ERROR_NOP);
+  if(err == NULL) return NULL;
```
MisterDA, line 585:
```suggestion
  if (err == NULL) return NULL;
```

flexdll.c
jmid: According to [this page](https://learn.microsoft.com/en-us/windows/win32/api/winnt/nf-winnt-interlockedcompareexchangepointer) the first destination parameter to `InterlockedCompareExchangePointer` should be `volatile`

flexdll.c
```diff
@@ -561,9 +579,20 @@ void flexdll_dlclose(void *u) {
...
 void *flexdll_dlsym(void *u, const char *name) {
-  if (u == &main_unit) return find_symbol_global(NULL,name);
-  else if (NULL == u) return find_symbol(&static_symtable,name);
-  else return find_symbol(((dlunit*)u)->symtbl,name);
+  void *res;
+  err_t * err;
+  err = get_tls_error(TLS_ERROR_NOP);
+  if (err == NULL) return NULL;
+
+  if (WaitForSingleObject(units_mutex, INFINITE) == WAIT_OBJECT_0) {
```
MisterDA, line 587:
```suggestion
  if (WaitForSingleObject(units_mutex, INFINITE) != WAIT_OBJECT_0) {
```
ouch, sorry!

jmid: No problem! I didn't notice in the CI - but found out locally when tests weren't passing :sweat_smile:

flexdll.c
jmid: I would think so. However I must either (a) be missing something or (b) messing up, because with that change `tests/lib-dynlink-domains` seems to loop infinitely :grimacing:

flexdll.c
jmid: Could it have to do with parallel access, e.g., in the presence of races to initialize it?

flexdll.c
jmid: I ended up reverting this one.

`WAIT_ABANDONED` does not offer `GetLastError`-integration, so I started adding an explicit error message case for it - only to then hit it during testing.

I therefore suppose these two cases were written to handle both `WAIT_OBJECT_0` and `WAIT_ABANDONED` as success-cases (`WAIT_TIMEOUT` should not happen with `INFINITE`...)

flexdll.c
```diff
@@ -496,8 +498,22 @@ void *flexdll_wdlopen(const wchar_t *file, int mode) {
 #endif /* __STDC_SECURE_LIB__ >= 200411L*/
 #endif /* CYGWIN */
 
+again:
+  if (units_mutex == INVALID_HANDLE_VALUE) {
+    HANDLE hMutex = CreateMutex(NULL, TRUE, NULL);
```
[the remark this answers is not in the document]
jmid: Added in 67efa7d

flexdll.c
jmid: (same for this one)

flexdll.c
jmid: OK, after looking more into this I'm convinced (I'm no memory-model expert!).
What's important to ensure synchronization is the atomic update and memory fence.
Addressed in 47480bc

flexdll.c
jmid: I tried this suggestion afresh this morning - and still encounter a loop (and eventually a timeout).
Are you experiencing the same @MisterDA?

flexdll.c
```diff
@@ -496,8 +498,26 @@ void *flexdll_wdlopen(const wchar_t *file, int mode) {
...
+    if (InterlockedCompareExchangePointer(&units_mutex, hMutex, INVALID_HANDLE_VALUE) != INVALID_HANDLE_VALUE) {
+      CloseHandle(hMutex);
+      goto again;
+    }
+  } else {
+    if (WaitForSingleObject(units_mutex, INFINITE) == WAIT_FAILED) {
+      if (!err->code) err->code = 1;
+      return NULL;
+    }
+  }
```
MisterDA, line 517:
```suggestion
  } else if (WaitForSingleObject(units_mutex, INFINITE) == WAIT_FAILED) {
    if (!err->code) err->code = 1;
    return NULL;
  }
```

MisterDA (approved):
I've tested an finally understood why the code is correct, thanks Jan! I think this is good to go.
Minor suggestion still.

flexdll.c
jmid: After some off-line, collective head scratching with @MisterDA (who was also observing looping), we've established that the while-loop is not equivalent, as it would attempt a `WaitForSingleObject` after a successful `CreateMutex` and `InterlockedCompareExchangePointer` whereas the label-version doesn't.

flexdll.c
jmid: Done in a824e52

jmid:
I've addressed the last minor comment, added a CHANGES entry, and rebased on `master`.

MisterDA (approved):

flexdll.c
dra27: My reasoning I think is that WAIT_ABANDONED should be treated like success?

dra27:
Thank you both for your work on this, and sorry for taking quite so long to merge it!

PR #141, opened 2024-07-29
Support relocation kind 0003 and extend IMAGE_REL_ types

### LINKED ISSUE #29: Unsupported relocation kind 0003 for __GSHandlerCheck in libcamlrun.lib(intern.obj)
I'm getting this when building OCaml 4.03.0 for x64 with bootstrapped flexlink and Visual Studio 2015. The relocation kind is `IMAGE_REL_AMD64_ADDR32NB`. I have not yet investigated exactly what that means.

jonahbeckford:
Fixes #29 . 

Specifically adds support for:

- IMAGE_REL_AMD64_ADDR32NB. (This is the "relocation kind 0003").

- IMAGE_REL_I386_DIR32NB. (This is the x86 version of the above).

- IMAGE_REL_AMD64_REL32_5. (This is documented but weirdly not implemented).

Tested with my ucrt branch https://github.com/jonahbeckford/flexdll/tree/0.43%2Bucrt where relocation kind 0003 occurs often. It especially occurs in "normal" C libraries (ucvrt; `/MD`) that are linked into an OCaml executable. None of the tested code used `/GS-`.

References:
- https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#coff-relocations-object-only
- https://github.com/llvm/llvm-project/blob/279953f1da7d7d0e404638414deec5d1df291c7c/lld/COFF/Chunks.cpp#L113-L158

### DIFF
--- a/flexdll.c
+++ b/flexdll.c
@@ -366,7 +366,7 @@ static void relocate(resolver f, void *data, reloctbl *tbl, err_t *err) {
         err->code = 3;
         goto restore;
       }
-      *((UINT32*) ptr->addr) = s;
+      *((UINT32*) ptr->addr) = (INT32) s;
       break;
     case RELOC_32NB:
       s += *((INT32*) ptr -> addr);
@@ -375,7 +375,7 @@ static void relocate(resolver f, void *data, reloctbl *tbl, err_t *err) {
         err->code = 3;
         goto restore;
       }
-      *((UINT32*) ptr->addr) = s;
+      *((UINT32*) ptr->addr) = (INT32) s;
       break;
     default:
       fprintf(stderr, "flexdll: unknown relocation kind");
--- a/flexdll.c
+++ b/flexdll.c
@@ -26,11 +26,15 @@
 typedef long intnat;
 typedef unsigned long uintnat;
 
+/* RELOC_ constants except RELOC_DONE have ordinal values based on when
+   they were introduced to the code base. These ordinal values are
+   persisted in .obj files so do not re-use any existing ordinals. */
 #define RELOC_REL32     0x0001
 #define RELOC_ABS       0x0002
-#define RELOC_REL32_4   0x0003
 #define RELOC_REL32_1   0x0004
 #define RELOC_REL32_2   0x0005
+#define RELOC_REL32_3   0x0008
+#define RELOC_REL32_4   0x0003
 #define RELOC_REL32_5   0x0006
 #define RELOC_32NB      0x0007
 #define RELOC_DONE      0x0100
@@ -328,16 +332,6 @@ static void relocate(resolver f, void *data, reloctbl *tbl, err_t *err) {
       }
       *((UINT32*) ptr->addr) = (INT32) s;
       break;
-    case RELOC_REL32_4:
-      s -= (INT_PTR)(ptr -> addr) + 8;
-      s += *((INT32*) ptr -> addr);
-      if (s != (INT32) s) {
-        sprintf(err->message, "flexdll error: cannot relocate RELOC_REL32_4, target is too far: %p  %p",(void *)((UINT_PTR) s), (void *) ((UINT_PTR)(INT32) s));
-        err->code = 3;
-        goto restore;
-      }
-      *((UINT32*) ptr->addr) = (INT32) s;
-      break;
     case RELOC_REL32_1:
       s -= (INT_PTR)(ptr -> addr) + 5;
       s += *((INT32*) ptr -> addr);
@@ -353,6 +347,26 @@ static void relocate(resolver f, void *data, reloctbl *tbl, err_t *err) {
       s += *((INT32*) ptr -> addr);
       if (s != (INT32) s) {
         sprintf(err->message, "flexdll error: cannot relocate RELOC_REL32_2, target is too far: %p  %p",(void *)((UINT_PTR) s), (void *) ((UINT_PTR)(INT32) s));
+        err->code = 3;
+        goto restore;
+      }
+      *((UINT32*) ptr->addr) = (INT32) s;
+      break;
+    case RELOC_REL32_3:
+      s -= (INT_PTR)(ptr -> addr) + 7;
+      s += *((INT32*) ptr -> addr);
+      if (s != (INT32) s) {
+        sprintf(err->message, "flexdll error: cannot relocate RELOC_REL32_3, target is too far: %p  %p",(void *)((UINT_PTR) s), (void *) ((UINT_PTR)(INT32) s));
+        err->code = 3;
+        goto restore;
+      }
+      *((UINT32*) ptr->addr) = (INT32) s;
+      break;
+    case RELOC_REL32_4:
+      s -= (INT_PTR)(ptr -> addr) + 8;
+      s += *((INT32*) ptr -> addr);
+      if (s != (INT32) s) {
+        sprintf(err->message, "flexdll error: cannot relocate RELOC_REL32_4, target is too far: %p  %p",(void *)((UINT_PTR) s), (void *) ((UINT_PTR)(INT32) s));
         err->code = 3;
         goto restore;
       }
--- a/flexdll.c
+++ b/flexdll.c
@@ -31,6 +31,8 @@ typedef unsigned long uintnat;
 #define RELOC_REL32_4   0x0003
 #define RELOC_REL32_1   0x0004
 #define RELOC_REL32_2   0x0005
+#define RELOC_REL32_5   0x0006
+#define RELOC_32NB      0x0007
 #define RELOC_DONE      0x0100
 
 typedef struct { UINT_PTR kind; char *name; UINT_PTR *addr; } reloc_entry;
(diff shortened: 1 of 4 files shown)


flexdll.c
```diff
@@ -356,6 +358,25 @@ static void relocate(resolver f, void *data, reloctbl *tbl, err_t *err) {
...
+      *((UINT32*) ptr->addr) = s;
+      break;
+    case RELOC_32NB:
+      s += *((INT32*) ptr -> addr);
+      if (s != (INT32) s) {
+        sprintf(err->message, "flexdll error: cannot relocate %s RELOC_32NB, target is too far: %p  %p", ptr->name, (void *)((UINT_PTR) s), (void *) ((UINT_PTR)(INT32) s));
+        err->code = 3;
+        goto restore;
+      }
+      *((UINT32*) ptr->addr) = s;
```
nojb, line 378: Just for my own understanding:

- In LLVM, this relocation is implemented by `add32(off, s)`, so that `off` in LLVM corresponds to `ptr + addr` in Flexlink and `s` in LLVM corresponds to `s` in Flexlink.
- The relocation `IMAGE_REL_AMD64_ADDR64` in LLVM (`RELOC_ABS` in Flexlink) is implemented by `add64(off, s + imageBase)`, so that `s + imageBase` in LLVM corresponds to `s` in Flexlink.

Do you understand why the `s` arguments do not seem to match between the two cases?

jonahbeckford: Going one step further, those equations imply that `imageBase = 0`.

I spent a lot of time looking and I couldn't find out why IMAGE_REL_AMD64_ADDR64 seems to work. The math for all the other relocation kinds make sense.

However, two things ...

1. The [`/base:` base address](https://learn.microsoft.com/en-us/cpp/build/reference/base-base-address?view=msvc-170) `link.exe` option is fixed for OCaml executable to be `0x10000` at https://github.com/ocaml/flexdll/blob/80496b51b9be9e128c6144377d20ccf6eff10129/cmdline.ml#L47
and also the same in
https://github.com/ocaml/flexdll/blob/80496b51b9be9e128c6144377d20ccf6eff10129/create_dll.ml#L77
for OCaml plugins / stubs.
2. All of the relative relocations from this section of code are translated by:
https://github.com/ocaml/flexdll/blob/80496b51b9be9e128c6144377d20ccf6eff10129/reloc.ml#L470-L475
into absolute relocations by `Reloc.abs`: https://github.com/ocaml/flexdll/blob/80496b51b9be9e128c6144377d20ccf6eff10129/coff.ml#L437-L444

So my suspicion was that the `Reloc.abs` absolute relocations were being translated at CreateProcess time by `ntdll.LdrInitializeThunk` (or whatever is reading the PE `.reloc` section) to complete the `imageBase` adjustment.

@dra27, any chance you can explain why `base_addr` (or `image_base` for DLLs) is not used in the calculations?

(These calculations are undocumented in flexdll and fairly complicated)

jonahbeckford:
I just noticed that I have not extended the DLL logic at https://github.com/ocaml/flexdll/blob/80496b51b9be9e128c6144377d20ccf6eff10129/create_dll.ml#L235-L253

That DLL logic *does* use the `image_base` for `IMAGE_REL_AMD64_ADDR64` so at least in that code the `IMAGE_REL_AMD64_ADDR64` math makes sense.

I don't really have a way to test that. The stubs/plugins generated by OCaml do not have those relocations (which is also why I didn't notice them).

flexdll.c
```diff
@@ -356,6 +358,25 @@ static void relocate(resolver f, void *data, reloctbl *tbl, err_t *err) {
...
       break;
+    case RELOC_REL32_5:
+      s -= (INT_PTR)(ptr -> addr) + 9;
+      s += *((INT32*) ptr -> addr);
+      if (s != (INT32) s) {
+        sprintf(err->message, "flexdll error: cannot relocate RELOC_REL32_5, target is too far: %p  %p",(void *)((UINT_PTR) s), (void *) ((UINT_PTR)(INT32) s));
+        err->code = 3;
+        goto restore;
+      }
+      *((UINT32*) ptr->addr) = s;
```
MisterDA, line 369:
```suggestion
      *((UINT32*) ptr->addr) = (INT32) s;
```

```
flexdll.c(369): warning C4244: '=': conversion from 'INT_PTR' to 'UINT32', possible loss of data
```

flexdll.c
MisterDA, line 378:
```suggestion
      *((UINT32*) ptr->addr) = (INT32) s;
```

```
flexdll.c(378): warning C4244: '=': conversion from 'INT_PTR' to 'UINT32', possible loss of data
```

dra27: It's far too long since I stared at all this, but are we comparing like with like here? Isn't that code in LLVM the part of a _linker_ which is computing actual relocations to be carried out by the _loader_? In flexdll's, case, I don't think we have to worry about the image base - the symbol table as a data structure has already been had all the relocations resolved when the executable was loaded (by the Windows loader) - so `s` in flexdll is an absolute address for the symbol, where in lld it's still base-relative?

reloc.ml
```diff
@@ -433,13 +433,18 @@ let add_reloc_table obj obj_name p =
...
+            0x0007 (* 32nb *)
+
         | `x64, 0x04 (* IMAGE_REL_AMD64_REL32 *)
         | `x86, 0x14 (* IMAGE_REL_I386_REL32 *) when not !no_rel_relocs ->
             0x0001 (* rel32 *)
 
         | `x64, 0x05 when not !no_rel_relocs -> 0x0004 (* rel32_1 *)
         | `x64, 0x08 when not !no_rel_relocs-> 0x0003 (* rel32_4 *)
         | `x64, 0x06 when not !no_rel_relocs-> 0x0005 (* rel32_2 *)
+        | `x64, 0x09 when not !no_rel_relocs-> 0x0006 (* rel32_5 *)
```
dra27, line 447: While we here, might we as well add the `rel32_3` case? I guess that these were included "on demand" in the past, but it does just look to be a set of them!

dra27 (commented):
This looks correct to me - at least, it sounds as though it's reasonably clear from your testing that the `RELOC_32NB` implementation is working, and the `RELOC_REL32_5` implementation is clearly correct w.r.t. the others already there.

Do you mind adding the last missing `RELOC_REL32_3` while we here, and I think this can go straight in.

flexdll.c
jonahbeckford: Yes, your loader comment is what I had meant with "translated at CreateProcess time by `ntdll.LdrInitializeThunk`".  But I am still puzzled why that only applies to IMAGE_REL_AMD64_ADDR64.

jonahbeckford:
This PR needs to stay open ... I am using this PR much more extensively and I'm encountering:

```
Fatal error: cannot load shared library dllthreads
Reason: flexdll error: cannot relocate caml_stat_free RELOC_REL32, target is too far: FFFFFFFF43282860  0000000043282860
```

```text
Fatal error: cannot load shared library dllcurl_stubs
Reason: flexdll error: cannot relocate Caml_state RELOC_REL32, target is too far: FFFFFFFF2DAA3206  000000002DAA3206
```

etc.

@MisterDA: Looks like casting of int64/uint64 to uint32 is not done correctly ... I think it was just luck before if it was working. Your [Avoid conversion from 'INT_PTR' to 'UINT32' warnings](https://github.com/ocaml/flexdll/pull/141/commits/b346f56bf8ae84300934d45a16d1c922c9b4db7a) review suggestion is very close to where I'm seeing the failures, so it sounds like the C compiler was giving us a good warning. If you know how to make those INT_PTR/INT32/UINT32 operations robust, reply back! I'll investigate more in the meantime.

reloc.ml
jonahbeckford: Yep. Will do after I get the existing relocations working properly! (My latest comment has the details).

dra27:
Indeed, the docs in flexdll are out of date w.r.t. `__declspec(dllimport)`. FlexDLL is showing its age, in that when it was first written Windows was still overwhelmingly 32-bit! Originally, the magic was that by specifying a base address for the DLL by default simply caused the 64-bit to load the DLLs close - there was never a guarantee, it just worked reliably for quite some time.

Except that we lost Cygwin64 support in 2018 with the release of Cygwin's `rebase` 4.4.4 (see [this thread](https://sourceware.org/pipermail/cygwin/2018-February/235936.html) for the explanation of why that change was made). That killed shared library / natdynlink support for Cygwin until 4.12 (and https://github.com/ocaml/ocaml/pull/9927). That's when the `__declspec(dllimport)` mechanism was restored (and is when the README ought to have been updated).

IIRC mingw-w64 had never clamped the base address because it hadn't been necessary, but that started to become necessary with binutils 2.36 - at which point (https://github.com/ocaml/ocaml/pull/10351) co-opted the same mechanism for mingw-w64. Hwoever, msvc64 still uses the base address trick.

dra27:
Are those failures with msvc64 or mingw-w64, @jonahbeckford? IIRC opam-repository-mingw "fixed" the binutils thing by forcing the base address for mingw-w64 instead, which means that this will be becoming a gradually bigger problem, requiring more uses of `CAMLextern` in libraries, indeed.

The systhreads failure is surprising, as I thought that was only in 5.1+ (and is fixed in 5.2.1 and 5.3.0) - which version of OCaml is that one with?

jonahbeckford:
> Are those failures with msvc64 or mingw-w64, @jonahbeckford? IIRC opam-repository-mingw "fixed" the binutils thing by forcing the base address for mingw-w64 instead, which means that this will be becoming a gradually bigger problem, requiring more uses of `CAMLextern` in libraries, indeed.

msvc64

> The systhreads failure is surprising, as I thought that was only in 5.1+ (and is fixed in 5.2.1 and 5.3.0) - which version of OCaml is that one with?

4.14.2 with the nine (9) ocaml-common-4_14-a* patches at https://github.com/diskuv/dkml-compiler/tree/main/src/p and two (2) non-DkML patches (see below) using the ucrt branch of flexdll (ucrt link in this PR's main comment).

ocaml-common-4_14-z02: This is relevant to systhreads. 
 https://gist.github.com/jonahbeckford/6cdbfb559b3f2c0b4eb143210bed0d3f. Of course, that is orthogonal to this PR but I'm curious what needed to be fixed with it. (I only backport a change when I see a problem, but I generally don't monitor the OCaml 5 changelog. I was expecting bugfixes to be backported to the 4.14 "LTS" version but that hasn't been happening.)

jonahbeckford:
> .... FlexDLL is showing its age

Thanks for the context! I feel at some point I might explore getting rid of flexdll entirely. There is a lot of complexity in flexdll for a a) use case centered around resolving circular references for plugins which I think was not a great design, and b) to remove the need for `__declspec(dll{im,ex}port)` which is no longer true today for 64-bit. Combine that with the sad state of flexdll security like missing ASLR and the (huge) effort I needed to remove `/GS-` from https://github.com/ocaml/flexdll/blob/a003c18c563ec661d29a304c0b7fb61910ac88d8/Makefile#L65

Basically, I think today there is little benefit and a lot of drawbacks where 10-20 years ago the balance was a lot more positive. But OCaml is deeply tied to the circular reference model for stubs and also (maybe related) deeply tied to a flat symbol namespace ... the two assumptions from ELF linking where DLL and dylib linking have better designs ... so I'm under no illusion that untangling the assumptions would be easy.

PR #146, opened 2024-11-21
Ensure the entrypoint is taken into account when computing the export table

dra27:
Presently on trunk and - as far as I can tell - since OCaml 3.11, in Cygwin64:

```console
$ make natruntop
OCaml version 5.4.0+dev0-2024-08-25 - native toplevel
Enter #help;; for help.

# #load "unix.cmxs";;
Error while loading otherlibs/unix/unix.cmxs: error loading shared library: Failure("Cannot resolve caml_debugger_in_use").
```
This is strange, since https://github.com/ocaml/ocaml/commit/27780d9f8a6812ac8d79c80565ebd5d991db8f65#diff-3c8eb806a6be5e7baa5af9546b7e12cdfa1220daae562235a450f87b93dd06e7 in OCaml 3.12 ensures that `debugger.o` is linked into OCaml programs precisely to prevent this problem. The symbol is indeed present:

```console
$ nm ocamlnat.exe | fgrep caml_debugger_in_use
0000000100f89818 B caml_debugger_in_use
```

the issue is that it is not included in the flexlink export table:

```console
$ PATH=opt/bin:$PATH ./boot/ocamlrun.exe ./ocamlopt.exe -nostdlib -I ./stdlib -I otherlibs/dynlink -I utils -I parsing -I typing -I bytecomp -I file_formats -I lambda -I middle_end -I middle_end/closure -I middle_end/flambda -I middle_end/flambda/base_types -I asmcomp -I driver -I toplevel -I tools -I runtime -I otherlibs/dynlink -I otherlibs/str -I otherlibs/systhreads -I otherlibs/unix -I otherlibs/runtime_events -g -linkall -I toplevel/native   -o ocamlnat.exe compilerlibs/ocamlcommon.cmxa compilerlibs/ocamloptcomp.cmxa compilerlibs/ocamlbytecomp.cmxa otherlibs/dynlink/dynlink.cmxa compilerlibs/ocamltoplevel.cmxa  toplevel/topstart.cmx -cclib -show-exports | fgrep caml_debugger_in_use
```

The reason is that while `-lasmrun` is passed to flexlink, the _objects_ passed on the command line give flexlink no reason to link `startup_nat.n.o`, which would then cause `debugger.n.o` to be considered and consequently these symbols to be added to the export table.

The fundamental fix here is that when linking an executable, flexlink now considers the entry point when closing over the required list of objects. For Cygwin, this fix is quite simple as it suffices to consider linking `main`. Cygwin's gcc does link [`crt0.o`](https://en.wikipedia.org/wiki/Crt0), but the full mechanism is technically a bit more involved. This doesn't matter, though, because being a Posix system, the entrypoint is always `main` (not `wmain` or `WinMain`/`wWinMain`, etc.).

For both MSVC and mingw-w64, the fix is a little more complex as, being non-Posix, there's not a strict requirement for the entrypoint to be `main`. For mingw-w64, the entrypoint is `mainCRTStartup`, but whether that calls `main` or `wmain` is determined by `-municode` which selects between using `crt2.o` or `crt2u.o` as the startup object. MSVC is affected both by the [`/ENTRY`](https://learn.microsoft.com/en-us/cpp/build/reference/entry-entry-point-symbol?view=msvc-170) and [`/SUBSYSTEM`](https://learn.microsoft.com/en-us/cpp/build/reference/subsystem-specify-subsystem?view=msvc-170) linker options, which select between one of `mainCRTStartup`, `wmainCRTStartup`, `WinMainCRTStartup` or `wWinMainCRTStartup`. `MSVCRT.lib` contains each of those symbols in a separate object each of which then has an import reference to either `main`, `wmain`, `WinMain` or `wWinMain` as required. Since OCaml 4.06, OCaml therefore uses `-municode` for mingw-w64 and `/ENTRY:wMainCRTStartup` for MSVC.

This leads to the main part of this PR - the entrypoint for mingw-w64 and MSVC is therefore defined not in one of the objects or libraries passed on the command line (i.e. in `libasmrun.a`), but in one of the default libraries (either `crt2u.o` or `MSVCRT.lib`). The problem is that while these libraries are scanned, they are not analysed when closing over the objects which need to be linked. This is correct (the default libraries are assumed to be being linked in both executable and DLL) - but the objects should be considered in case they import symbols which _do_ come from the objects and libraries passed on the command line. The changes here therefore mean that for mingw-w64, `mainCRTStartup` gets resolved to `crt2u.o`. That module imports `wmain` which then causes `main.n.o` from `libasmrun.a` to be linked and, consequently `startup_nat.n.o` and hence `debugger.n.o`. This PR happens to be an alternate fix (or, rather, workaround) for https://github.com/ocaml/ocaml/pull/13520.


reloc.ml
```diff
@@ -1007,16 +1013,75 @@ let build_dll link_exe output_file files exts extra_args =
...
+  let entrypoint =
+    if not main_pgm then
+      None
+    else
+      match !toolchain with
+      | `CYGWIN64 ->
+          Some "main"
+      | `MINGW | `MINGW64 -> begin
+          let entry_point s =
+            String.length s > 7 && String.sub s 0 7 = "-Wl,-e,"
```
nojb, line 1037: Isn't `-Wl,-e` rewritten into `-link` arguments (cf https://github.com/ocaml/flexdll/blob/ab8161fc7161a598c5aaf66f02298035639d5ef3/cmdline.ml#L281-L285) ?

nojb (approved):
I only glanced at the code, but LGTM

reloc.ml
dra27: Urgh - good catch, and unfortunately it's because this part is completely broken. The problem here is that when we link MSVC we actually call the linker, but when we link mingw-w64, we call the C compiler! I added this bit here for completeness - for MSVC, we of course select Unicode with `-link /entry:wmainCRTStartup` and that works because flexlink passes the `/entry:` directly to `/link`. For mingw-w64, we of course do `-link -municode` and that survives being passed to the C compiler to do the linking, but `-Wl,-e,foo` won't work being passed to the C compiler as `-e foo`. I only added this handling for `-Wl,-e` because I just written the `/entry:` part for MSVC (which I knew was needed).

Hmm. Given that you can't specify the entry point at the moment on mingw-w64, I'm tempted to remove this bit of code and just add a comment that you can't change the entrypoint on mingw-w64?

nojb: Sounds reasonable!

dra27: Oh - coffee kicks in! I remember how this works - the `-Wl,` translation to `-link` is a shim to allow common options to pass through. You _can_ specify the entry point on mingw-w64, it's just that you _have_ to say `-link -Wl,-e,entrypoint` at which point the code above _does_ work as intended!

CHANGES
```diff
@@ -37,6 +37,9 @@ Version 0.44
   as querying the compilers for their library search paths. Support parsing
   clang output.
   (Antonin Décimo, review by David Allsopp)
+- GPR#146: Take the entrypoint (main, wmainCRTStartup, etc.) into account when
+  determining the modules which will be linked. For mingw-w64, select crt2u.o
+  instead of crt2.o if -link -municode is specified (David Allsopp)
```
MisterDA, line 42: This should be moved in the section above.

MisterDA (approved):

reloc.ml
```diff
@@ -1423,8 +1488,14 @@ let setup_toolchain () =
...
     default_libs := "-lmingw32" :: "-lgcc" :: !default_libs;
-    if !exe_mode = `EXE then default_libs := "crt2.o" :: !default_libs
-    else default_libs := "dllcrt2.o" :: !default_libs
+    if not (List.mem "-nostartfiles" !extra_args) then begin
+      if !exe_mode = `EXE then
+        if List.mem "-municode" !extra_args then
+          default_libs := "crt2u.o" :: !default_libs
+        else
+          default_libs := "crt2.o" :: !default_libs
+      else default_libs := "dllcrt2.o" :: !default_libs
```
MisterDA, line 1497:
```suggestion
      let crt =
        if !exe_mode = `EXE then
          if List.mem "-municode" !extra_args then "crt2u.o" else "crt2.o"
        else "dllcrt2.o"
      in
      default_libs := crt :: !default_libs
```

A bit terser, perhaps?

PR #147, opened 2024-11-22
Make `-lgcc_eh` on mingw-w64 optional

### LINKED ISSUE #133: Conditionally link with gcc_eh
Linking with `gcc_eh` should be conditionally done when it is available. See https://github.com/ocaml/ocaml/pull/12996 for a similar discussion for the ocaml compiler.

dra27:
Should fix #133. As noted in ocaml/ocaml#12996, there are configurations where `-lgcc_eh` isn't available (and therefore required). The change is simply to resolve `-lgcc_eh` and only add it to `default_libs` if it's actually found - although it's a slightly noisier change than that simplicity implies!

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -20,6 +20,7 @@ Next version
 - GPR#156: Allow choosing an alternative linker and manifest tool at runtime
   with -use-linker=<cmd> and -use-mt=<cmd>.
   (Antonin Décimo, review by David Allsopp)
+- GPR#133, GPR#147: Allow default libraries to be marked as optional (David Allsopp, report by Romain Beauxis)
 
 Version 0.43
 - GPR#108: Add -lgcc_s to Cygwin's link libraries, upstreaming a patch from the
--- a/reloc.ml
+++ b/reloc.ml
@@ -317,7 +317,7 @@ let rec map_until_found f = function
       | r ->
           r
 
-let find_file =
+let find_file_exn =
   let memo = Hashtbl.create 16 in
   fun fn ->
     let k = String.lowercase_ascii fn in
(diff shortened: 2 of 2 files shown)


dra27:
@toots - no particular rush, but if you have a chance to confirm that this allows the patch you carry in opam-cross-windows to be removed, that'd be great.

toots:
Hi @dra27. The changes in https://github.com/ocaml/ocaml/pull/12996 are actually the patch used by the cross-compiler so it can already be dropped.

I'm not sure that I understand the difference with this PR or what additional change this PR is doing? `gcc_eh` was still mentioned in the compilation workflow somehow?

dra27:
Ah, have I misunderstood the problem - I thought the issue was that there wasn't any `libgcc_eh.a` at all, so `-lgcc_eh` was failing. The change here is that it tries to see if there is an `-lgcc_eh`. Am I missing something about the failure mode that was being detected before?

dra27:
Isn't this patch https://github.com/ocaml-cross/opam-cross-windows/blob/main/packages/flexdll-windows/flexdll-windows.0.42/files/patches/no-gcc_eh.patch still in use, though?

toots:
> Isn't this patch https://github.com/ocaml-cross/opam-cross-windows/blob/main/packages/flexdll-windows/flexdll-windows.0.42/files/patches/no-gcc_eh.patch still in use, though?

Ho, different one! I was looking at the patch on OCaml: https://github.com/ocaml-cross/opam-cross-windows/blob/main/packages/ocaml-windows64/ocaml-windows64.5.1.1/files/patches/configure.patch

Then, yes, this looks like it will allow us to remove the patch on flexdll. I should be able to test this soon, thanks!

dra27:
Have you been able to look at this patch, @toots?

toots:
I think so sorry I forgot to respond. This looks good to me.

toots (approved):

Issue #142, opened 2024-08-20
doesn't recognize -pthread option


ygrek:
e.g. https://github.com/ygrek/mldonkey/actions/runs/10475349585/job/29011798015#step:7:26

I am pretty sure I don't need this option for ocaml builds and can find workarounds, but just to let you know

ftr from `man gcc`: 
```
       -pthread
           Define additional macros required for using the POSIX threads library.  You should use this option consistently for both compilation and
           linking.  This option is supported on GNU/Linux targets, most other Unix derivatives, and also on x86 Cygwin and MinGW targets.
```

dra27:
Note that for OCaml 5.x this flag should be unnecessary on all builds, as the runtime requires pthreads (in fact, I think adding `-pthread` might even cause problems for the mingw-w64 build, because it links pthreads statically).

For mingw-w64 on OCaml 4.x - but the key thing is that it should be being added as `-link -pthread` (i.e. it should be being passed from OCaml as `-ccopt -link -ccopt -pthread`)

dra27:
This is fixed for gcc-based toolchains (i.e. for mingw-w64) by passing the option on directly - note that if compiling for the MSVC port of OCaml, there'll still need to be handling on the caller's side to avoid passing `-pthread` at all.

PR #148, opened 2024-11-30
Pass -pthread to the linker

toots:
The `-pthread` option keeps surfacing through the different compilation configurations I see when cross-compiling.

This PR makes it so the option is accepted and passed down to the linker.

### DIFF
--- a/cmdline.ml
+++ b/cmdline.ml
@@ -89,8 +89,8 @@ let specs = [
   "-base", Arg.String (fun s -> base_addr := s),
   " Specify base address (Win64 only)";
 
-  "-pthread", Arg.Unit (fun () -> ()),
-  "Ignored";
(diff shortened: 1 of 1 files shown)


dra27:
cf. #142 - I'm not (yet) sure quietly dropping this is the "right" thing to do. Would it be better to be recognising it and passing it on instead, like the handling for `-Wl,`?

toots:
I think passing it down to the linker is always better as it will mimick what the original call would expect. I ignored it because I believe it is ignored by the linker but it is always safer to not make this assumption and pass it down.

toots:
@dra27 I believe the changes should be passing the option down to the linker now.

dra27 (approved):
LGTM, thanks!

PR #150, opened 2025-01-14
Fix stack reserve to read from the COFF file header

MisterDA:
An invocation of flexlink (with some debug patches added) would raise:

```console
> flexlink -exe -chain msvc64 -merge-manifest -stack 33554432 -link /ENTRY:wmainCRTStartup  -o runtime/ocamlrun.exe runtime/prims.obj runtime/libcamlrun.lib ws2_32.lib ole32.lib uuid.lib advapi32.lib shell32.lib version.lib shlwapi.lib synchronization.lib
Raised by primitive operation at Coff.int16 in file "coff.ml", line 242, characters 30-49
Called from Coff.Stacksize.set_stack_reserve in file "coff.ml", line 1055, characters 12-26
Called from Reloc.patch_output in file "reloc.ml", line 654, characters 16-54
Cannot set stack reserve: Invalid_argument("index out of bounds")
```

After investigating, I've found out that flexlink incorrectly reads the PE executable. The [COFF File Header (Object and Image)][1] starts just after the signature `"PE\0\0"`. The `read ic 0 20` seeks the offset 0 (the start) in the PE file, instead of the offset of the header. This leads to incorrectly reading ~and writing~ values in the PE headers.

[1]: https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#coff-file-header-object-and-image

The first patch fixes the problem, the second renames the variables to follow the documentation more closely.

### DIFF
--- a/coff.ml
+++ b/coff.ml
@@ -1044,10 +1044,11 @@ module Stacksize = struct
       else filename
     in
     let ic = open_in_bin filename in
-    let hdr_offset = int16 (read ic 0x3c 2) 0 in
+    (* Read e_lfanew from the IMAGE_DOS_HEADER (see winnt.h) *)
+    let hdr_offset = int32_ (read ic 0x3c 4) 0 in
     let pe_signature = read_str ic hdr_offset 4 in
     assert(pe_signature = "PE\000\000");
-    let coff_hdr = read ic 0 20 in
+    let coff_hdr = read ic (hdr_offset + 4) 20 in
     let opthdr_size = int16 coff_hdr 16 in
     let opthdr = read ic (hdr_offset + 24) opthdr_size in
     let machine =
@@ -1057,7 +1058,6 @@ module Stacksize = struct
       | magic -> Printf.ksprintf failwith "Cannot determine image target (magic = %x)." magic
     in
     let reserve_offset = hdr_offset + 24 + 72 in
-(*    Printf.printf "current stack reserve %ld\n%!" (int32 opthdr 72); *)
     close_in ic;
 
     let oc = open_out_gen [Open_wronly; Open_binary] 0x777 filename in
--- a/coff.ml
+++ b/coff.ml
@@ -1038,31 +1038,31 @@ end
 
 
 module Stacksize = struct
-  let set_stack_reserve filename reserve =
+  let set_stack_reserve filename size_of_stack_reserve =
     let filename =
       if not (Sys.file_exists filename) && (Sys.file_exists (filename ^ ".exe")) then filename ^ ".exe"
       else filename
     in
     let ic = open_in_bin filename in
     (* Read e_lfanew from the IMAGE_DOS_HEADER (see winnt.h) *)
-    let hdr_offset = int32_ (read ic 0x3c 4) 0 in
-    let pe_signature = read_str ic hdr_offset 4 in
+    let pe_signature_offset = int32_ (read ic 0x3c 4) 0 in
+    let pe_signature = read_str ic pe_signature_offset 4 in
     assert(pe_signature = "PE\000\000");
-    let coff_hdr = read ic (hdr_offset + 4) 20 in
-    let opthdr_size = int16 coff_hdr 16 in
-    let opthdr = read ic (hdr_offset + 24) opthdr_size in
-    let machine =
-      match int16 opthdr 0 with
-      | 0x10b -> `x86
-      | 0x20b -> `x64
-      | magic -> Printf.ksprintf failwith "Cannot determine image target (magic = %x)." magic
+    let coff_file_header = read ic (pe_signature_offset + 4) 20 in
+    let size_of_optional_header = int16 coff_file_header 16 in
+    let optional_header = read ic (pe_signature_offset + 4 + 20) size_of_optional_header in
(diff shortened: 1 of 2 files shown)


coff.ml
```diff
@@ -1047,7 +1047,7 @@ module Stacksize = struct
     let hdr_offset = int16 (read ic 0x3c 2) 0 in
```
nojb: This is orthogonal to your patch, but isn't the `hdr_offset` 32-bit wide? https://en.wikipedia.org/wiki/Portable_Executable

nojb (approved):
Personally I would have left out the second commit :), but I'm not against it.

LGTM

coff.ml
MisterDA: You're right, it's a `LONG`. I'll update the commit!

MisterDA:
How come the bug hasn't been noticed since 2012?…

coff.ml
```diff
@@ -1038,31 +1038,34 @@ end
...
-  let set_stack_reserve filename reserve =
+  let set_stack_reserve filename size_of_stack_reserve =
     let filename =
       if not (Sys.file_exists filename) && (Sys.file_exists (filename ^ ".exe")) then filename ^ ".exe"
       else filename
     in
     let ic = open_in_bin filename in
-    let hdr_offset = int16 (read ic 0x3c 2) 0 in
-    let pe_signature = read_str ic hdr_offset 4 in
+    let pe_signature_offset = int32 (read ic 0x3c 4) 0 |> Int32.to_int in
```
MisterDA, line 1047: If the offset fits in 16 bits, I suppose it's safe to forget the high-order bit from int32 to int.

coff.ml
MisterDA: If the offset fits in 16 bits, I suppose it's safe to forget the high-order bit from int32 to int.

coff.ml
dra27, line 1047: We have an aptly named function for that, and we can add a primary reference to the field (given that the PE spec doesn't state its size):
```suggestion
    (* Read e_lfanew from the IMAGE_DOS_HEADER (see winnt.h) *)
    let pe_signature_offset = int32_ (read ic 0x3c 4) 0 in
```

dra27 (approved):
A tiny tweak to suggest, but otherwise thanks for this!

It's naturally great to fix this, but just to note that I don't think there was ever a problem with _writing_ executables. The bug caused the wrong header size to be _read_ but the optional header was being accessed at the correct offsets (and it was only be read to determine if it's PE32 or PE32+) - i.e. as long as the incorrectly read size was not less than 2 or greater than the size of the file, then the code worked correctly (and wrote the correctly sized 32 or 64bit value to the correct place in the executable).

While I started being a little grumpy about the renaming, actually I think it's well worth having on this occasion, thank you! The name `machine` was especially confusing as there _is_ a field machine in the COFF header, but the field being read was actually the magic number from the Optional Header.

coff.ml
```diff
@@ -1038,31 +1038,35 @@ end
...
+      | 0x10b -> `PE32
+      | 0x20b -> `PE32plus
+      | magic -> Printf.ksprintf failwith "Cannot determine PE format (magic = %x)." magic
     in
-    let reserve_offset = hdr_offset + 24 + 72 in
-(*    Printf.printf "current stack reserve %ld\n%!" (int32 opthdr 72); *)
+    let size_of_stack_reserve_offset = pe_signature_offset + 4 + 20 + 72 in
+    (* begin match pe_format with
+    | `PE32 -> Printf.printf "SizeOfStackReserve 0x%08x\n%!" (int32 opthdr 72)
+    | `PE32plus -> Printf.printf "SizeOfStackReserve 0x%016Lx\n%!" (int64 opthdr 72)
```
MisterDA, line 1063: There is no `int64` function, but if there was, this would print the value of the field as hex (it's unsigned and little-endian). The bit twiddling is fastidious.

coff.ml
```diff
@@ -1038,31 +1038,35 @@ end
...
+    end; *)
     close_in ic;
 
     let oc = open_out_gen [Open_wronly; Open_binary] 0x777 filename in
-    seek_out oc reserve_offset;
-    emit_int32 oc reserve;
-    if machine = `x64 then emit_int32 oc 0l;
+    seek_out oc size_of_stack_reserve_offset;
+    emit_int32 oc size_of_stack_reserve;
+    if pe_format = `PE32plus then emit_int32 oc 0l;
```
MisterDA, line 1070: Should we expand this to support stack sizes bigger than $2^{32}-1$ bytes?

dra27: Unless there’s a call for it, no! I can’t find a reference to it, but I don’t think it’s supported even on 64bit Windows

coff.ml
dra27: I hadn't spotted this bit of the change the first time around - I'm not that fond of the commented out debugging bits in the codebase as it is, but I'm not sure that having commented out code which has _never_ worked is a positive change! Can I suggest we just delete the original comment (which would have been there from debugging the original 32-bit implementation, I expect)?

MisterDA: Done, I've removed it.

PR #151, opened 2025-01-25
Use response files with all the toolchains

punchagan:
Response files were added to gcc 4.2.0 [1], but Cygwin was stuck on gcc
3.4.4 for ages until mingw-w64 was introduced at gcc 4.5 (Cygwin's gcc
itself still lagged for a while) cf. [2]. (Based on inputs from @dra27).

So, in 2008 [3], it was correct to avoid using response files with
CYGWIN & MINGW. Later, support was added for the GNAT toolchain based on
the MINGW toolchain, also avoiding the use of response files. But, now
all the platforms could benefit from using response files. This commit
enables the use of response files for all toolchains, including GNAT
since the GNAT compiler is based on gcc [4].

The UTF-16 conversion of paths seems to have been primarily implemented
for MSVC[5] and causes issues with MINGW64. The linker fails with a
"file not found" error with a weird 2 character file path, which seems
like an encoding issue when reading the Byte Order Mark written during
the conversion. This commit turns off the UTF-16 conversion when not
using the MSVC or LIGHTLD toolchains.

[1] - https://github.com/gcc-mirror/gcc/commit/9d5305381ff11ba5615705672115079ffb7b0750
[2] - http://ctm.crouchingtigerhiddenfruitbat.org/pub/cygwin/circa/2010/09/14/020013/index.html
[3] -
https://github.com/ocaml/flexdll/commit/fb849384d6c73232c588cae0535bed63a3330798
[4] - https://github.com/ocaml/flexdll/pull/151#issuecomment-2613980304
[5] - https://github.com/ocaml/flexdll/issues/36

### DIFF
--- a/reloc.ml
+++ b/reloc.ml
@@ -183,29 +183,22 @@ let build_diversion lst =
         Bytes.to_string s ^ "\r\n"
       ) (List.filter (fun f -> f <> "") lst)
   in
-  let utf16, lst =
-    try true, List.map toutf16 lst
-    with Not_utf8 -> false, lst
+  let lst =
+    match !toolchain with
+    | `MINGW | `MINGW64 | `GNAT | `GNAT64 | `CYGWIN64 -> lst
+    | `MSVC | `MSVC64 | `LIGHTLD ->
+      (* UTF-16 response files required *)
+      try
+        let lst = List.map toutf16 lst in
+        output_string oc "\xFF\xFE"; (* LE BOM *)
+        lst
+      with Not_utf8 -> lst
   in
-  if utf16 then output_string oc "\xFF\xFE"; (* LE BOM *)
   List.iter (fun s -> output_string oc s) lst;
   close_out oc;
   "@" ^ responsefile
 
-type cmdline = {
-    may_use_response_file: bool;
-  }
-
-let new_cmdline () =
-  let rf = match !toolchain with
-  | `MSVC | `MSVC64 | `LIGHTLD -> true
-  | `MINGW | `MINGW64 | `GNAT | `GNAT64 | `CYGWIN64 -> false
-  in
-  {
-   may_use_response_file = rf;
-  }
-
-let run_command cmdline cmd =
+let run_command cmd =
   let pipe_to_null = (!toolchain = `MSVC || !toolchain = `MSVC64) in
   let silencer = if pipe_to_null then " >NUL 2>NUL" else ""
   in
@@ -238,11 +231,11 @@ let run_command cmdline cmd =
       failwith "Error during linking\n"
     end
 
-let quote_files cmdline lst =
+let quote_files lst =
   let s =
     String.concat " "
       (List.map (fun f -> if f = "" then f else Filename.quote f) lst) in
-  if String.length s >= 1024 && cmdline.may_use_response_file then Filename.quote (build_diversion lst)
+  if String.length s >= 1024 then Filename.quote (build_diversion lst)
   else s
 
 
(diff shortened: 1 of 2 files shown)


yakobowski:
The support for GNAT has been added for AdaCore, butt I think we've actually stopped using this codepath entirely. Unfortunately I've left AdaCore recently, so you will need to double-check with e.g. @kanigsson.

In any case, GNAT is based on GCC, and is only lagging a few versions behind. Typically, the latest versions are GCC-13 based. So I would suggest using response files in all cases.

punchagan:
> In any case, GNAT is based on GCC, and is only lagging a few versions behind. Typically, the latest versions are GCC-13 based. So I would suggest using response files in all cases.

Thank you for your suggestion! I'll update the PR

punchagan:
> In any case, GNAT is based on GCC, and is only lagging a few versions behind. Typically, the latest versions are GCC-13 based. So I would suggest using response files in all cases.

Updated the PR to use toolchains for GNAT too!

reloc.ml
```diff
@@ -184,28 +184,16 @@ let build_diversion lst =
       ) (List.filter (fun f -> f <> "") lst)
   in
   let utf16, lst =
-    try true, List.map toutf16 lst
-    with Not_utf8 -> false, lst
+    if !toolchain = `MSVC || !toolchain = `MSVC64 || !toolchain = `LIGHTLD
+    then try (true, List.map toutf16 lst) with Not_utf8 -> (false, lst)
+    else (false, lst)
```
MisterDA, line 189: Maybe a match statement would look better here?

kanigsson:
Sorry I missed this mention by @yakobowski. GNAT behaves like GCC here.
Contrary to what he said, I believe we still use the GNAT-specific flexdll changes, so we would be grateful if they could be maintained. Thank you!

MisterDA (approved):

shonfeder:
Hello! I thought it was worth checking in here to see if we may be OK merging this, or if there additional validation we may do to build the needed confidence.

reloc.ml
```diff
@@ -184,28 +184,17 @@ let build_diversion lst =
       ) (List.filter (fun f -> f <> "") lst)
   in
   let utf16, lst =
```
dra27, line 186:
```suggestion
  let lst =
```

reloc.ml
```diff
@@ -184,28 +184,17 @@ let build_diversion lst =
...
-    try true, List.map toutf16 lst
-    with Not_utf8 -> false, lst
+    match !toolchain with
+    | `MSVC | `MSVC64 | `LIGHTLD ->
+        (try (true, List.map toutf16 lst) with Not_utf8 -> (false, lst))
+    | _ -> (false, lst)
   in
   if utf16 then output_string oc "\xFF\xFE"; (* LE BOM *)
```
dra27, line 192: I'd prefer to have the options fully enumerated (as was on L201-202) below, but overall I wonder if this would be clearer for the future:
```suggestion
    match !toolchain with
    | `MINGW | `MINGW64 | `GNAT | `GNAT64 | `CYGWIN64 -> lst
    | `MSVC | `MSVC64 | `LIGHTLD ->
        (* UTF-16 response files required *)
        try
          let lst = List.map toutf16 lst in
          output_string oc "\xFF\xFE"; (* LE BOM *)
          lst
        with Not_utf8 -> lst
  in
```

dra27 (commented):
Apologies for the time it's taken me to sit down with this - semantically, this is good to go, thank you.

Please could you push an update to the `CHANGES` file, and I'd prefer either the full change I suggest below or at least for the wildcard in the match to be replaced with the full expansion of the other linkers.

punchagan:
> Apologies for the time it's taken me to sit down with this - semantically, this is good to go, thank you.

No worries at all. Thanks for taking the time to review it!

> Please could you push an update to the `CHANGES` file, and I'd prefer either the full change I suggest below or at least for the wildcard in the match to be replaced with the full expansion of the other linkers.

I've incorporated your suggestions and updated the `CHANGES` file. Thanks!

dra27:
Thank you!

PR #152, opened 2025-01-31
Use a file to pass arguments to cygpath

punchagan:
When linking a fairly large OCaml project in Windows, I run into an 'argument list too long' error when trying to resolve filenames in `build_dll`. This commit switches to using a file to pass a long arguments list to cygpath, similar to our use of response files.

See also #151

### DIFF
--- a/reloc.ml
+++ b/reloc.ml
@@ -244,7 +244,17 @@ let quote_files lst =
 let cygpath l cont =
   let accept_error = (!use_cygpath = `Try && l <> []) in
   let l =
-    let args = String.concat " " (List.map Filename.quote l) in
+    let s = "-- " ^ String.concat " " (List.map Filename.quote l) in
+    let args =
+      if not Sys.cygwin && String.length s >= 1024 then begin
+        (* cygpath loads the file in "text" mode, so CRLF endings are fine *)
+        let (fn, oc) = open_temp_file "cygpathargs" "" in
+        List.iter (fun x -> output_string oc x; output_char oc '\n') l;
+        close_out oc;
+        "--file " ^ Filename.quote fn
+      end else
+        s
+    in
     get_output ~accept_error "cygpath -m %s" args
   in
   if accept_error && l = [] then begin
--- a/CHANGES
+++ b/CHANGES
@@ -16,6 +16,7 @@ Next version
   Fix FlexDLL cross-build on macOS and BSD.
   (Antonin Décimo, review by Nicolás Ojeda Bär)
 - GPR#151: Use response files with all toolchains (Puneeth Chaganti, review by David Allsopp and Antonin Décimo)
+- GPR#152: Use a file to pass arguments to cygpath (Puneeth Chaganti, review by David Allsopp and Antonin Décimo)
 
 Version 0.43
 - GPR#108: Add -lgcc_s to Cygwin's link libraries, upstreaming a patch from the
--- a/Compat.ml.in
+++ b/Compat.ml.in
@@ -136,6 +136,7 @@
 401:  include Sys
 401:
 401:  let win32 = (Sys.os_type = "Win32")
+401:  let cygwin = (Sys.os_type = "Cygwin")
 401:end
 
 402:type bytes = string


reloc.ml
```diff
@@ -251,7 +251,16 @@ let quote_files cmdline lst =
 let cygpath l cont =
   let accept_error = (!use_cygpath = `Try && l <> []) in
   let l =
-    let args = String.concat " " (List.map Filename.quote l) in
+    let s = String.concat " " (List.map Filename.quote l) in
+    let args =
+      if String.length s >= 1024 then begin
```
MisterDA, line 256: I suggest factoring this constant with the one in `quote_files`. Something like:

```ocaml
(* Use response files if the command line length exceeds 1024
   characters (friendlier for humans than 8192). *)
let max_command_line_length = 1024
```

MisterDA: My bad, there's already this:

```ocaml
(* This is the longest command which can be passed to [Sys.command] *)
let max_command_length =
  let processor = try Sys.getenv "COMSPEC" with Not_found -> "cmd.exe" in
  (* The 4 is from the " /c " *)
  8191 - String.length processor - 4
```

Let's keep 1024.

dra27, line 256: Cygwin isn't affected by command line length (different mechanism used), so out of an abundance of caution, can we have:
```suggestion
      if not Sys.cygwin && String.length s >= 1024 then begin
```

reloc.ml
```diff
@@ -251,7 +251,16 @@ let quote_files cmdline lst =
...
+        let (fn, oc) = open_temp_file "cygpathargs" "" in
+        List.iter (fun x -> output_string oc (x ^ "\n")) l;
```
dra27, line 258: FWIW, we may as well avoid all those allocations!
```suggestion
        List.iter (fun x -> output_string oc x; output_char oc '\n') l;
```

reloc.ml
```diff
@@ -251,7 +251,16 @@ let quote_files cmdline lst =
...
+        close_out oc;
+        "--file " ^ fn
```
dra27, line 260:
```suggestion
        "--file " ^ Filename.quote fn
```

reloc.ml
```diff
@@ -251,7 +251,16 @@ let quote_files cmdline lst =
 let cygpath l cont =
   let accept_error = (!use_cygpath = `Try && l <> []) in
   let l =
-    let args = String.concat " " (List.map Filename.quote l) in
+    let s = String.concat " " (List.map Filename.quote l) in
+    let args =
+      if String.length s >= 1024 then begin
+        let (fn, oc) = open_temp_file "cygpathargs" "" in
```
dra27, line 257: cf. https://github.com/cygwin/cygwin/blob/main/winsup/utils/cygpath.cc#L1081 - useful, because the `?perms` argument is 4.03+ 🫣
```suggestion
        (* cygpath loads the file in "text" mode, so CRLF endings are fine *)
        let (fn, oc) = open_temp_file "cygpathargs" "" in
```

reloc.ml
```diff
@@ -251,7 +251,16 @@ let quote_files cmdline lst =
 let cygpath l cont =
   let accept_error = (!use_cygpath = `Try && l <> []) in
   let l =
-    let args = String.concat " " (List.map Filename.quote l) in
+    let s = String.concat " " (List.map Filename.quote l) in
```
dra27, line 254: While we here, we could harden this:
```suggestion
    let s = "-- " ^ String.concat " " (List.map Filename.quote l) in
```

dra27 (commented):
As with #151, sorry also for the slow review! I think I'd prefer to gate this strictly for the native Windows build of `flexlink.exe` and the argument to `--file` should be quoted.

punchagan:
> As with #151, sorry also for the slow review! I think I'd prefer to gate this strictly for the native Windows build of `flexlink.exe` and the argument to `--file` should be quoted.

Thank you for taking the time to carefully review this PR, and your helpful suggestions! I've pushed changes with your suggestions and updated the `CHANGES` file.

reloc.ml
punchagan: Looks like `Sys.cygwin` is only `4.01+`.

punchagan: I changed the check to `Sys.os_type <> "Cygwin"`

punchagan: I've now added a compatibility shim. Thanks @MisterDA !

dra27 (approved):
Thank you!

PR #153, opened 2025-02-14
Makefile fixes

MisterDA:
This PR allows users to:
- pass `MSVC` and `MSVC64` variables to Make when building FlexDLL to allow choosing an alternative `cl.exe`-like compiler, say, `clang-cl`;
- pass `RC` to choose an alternative resource compiler (like `rc.exe` or `windres`), say `llvm-rc` or `llvm-windres`;
- pass `MINCC`, `MIN64CC`, `CYG64CC` to choose (respectively) an alternative MinGW-w64 (32 bits), MinGW-w64 (64 bits), Cygwin (64 bits) compiler. Their triplet prefix can be overridden with (respectively) `MINCC_PREFIX`, `MIN64CC_PREFIX`, `CYG64_PREFIX`.

It has also two Makefile fixes: one to support POSIX (really BSD) sed (looking at you, macOS!) and one to add missing OCaml versions when processing `Compat.ml.in` into `Compat.ml`.


nojb (commented):
Don't know about the `sed` change and have not tested, but otherwise LGTM

MisterDA:
Thanks. I've added support for `WINDRES` too, and a change entry.

The `sed` change makes it POSIX-compatible. [GNU sed manual](https://www.gnu.org/software/sed/manual/sed.html#Overview-of-basic-regular-expression-syntax) reads (emphasis mine):

> `regexp1\|regexp2`
> Matches either `regexp1` or `regexp2`. Use parentheses to use complex alternative regular expressions. The matching process tries each alternative in turn, from left to right, and the first one that succeeds is used. *It is a GNU extension*.

it also has this note:

> `^` acts as a special character only at the beginning of the regular expression or subexpression (that is, after `\(` or `\|`). Portable scripts should avoid `^` at the beginning of a subexpression, though, as POSIX allows implementations that treat `^` as an ordinary character in that context. 

so perhaps `^` in the whole expression should be factored out, but who cares.

MisterDA:
I think I've finally converged and haven't missed anything anymore.
The CI failure is a network failure.

MisterDA:
> and haven't missed anything anymore.

famous last words

nojb:
> > and haven't missed anything anymore.
> 
> famous last words

Are you confident on your testing @MisterDA? If yes, I would say we can move to merge.

Makefile
```diff
@@ -190,7 +190,7 @@ FLEXDLL_VS_VERSION_INFO = $(subst .,$(COMMA),$(FLEXDLL_FULL_VERSION))
 
 RC_FLAGS = \
   /d FLEXDLL_VS_VERSION_INFO=$(FLEXDLL_VS_VERSION_INFO) \
-  /d FLEXDLL_FULL_VERSION="$(FLEXDLL_FULL_VERSION)"
+  /d FLEXDLL_FULL_VERSION=\"$(FLEXDLL_FULL_VERSION)\"
```
dra27, line 193: Is this _definitely_ not breaking MSVC builds?

Makefile
```diff
@@ -28,14 +28,14 @@ endif
 
 GCC_FLAGS = -Wall
 
-MINGW_PREFIX = i686-w64-mingw32-
-MINCC = $(MINGW_PREFIX)gcc $(GCC_FLAGS)
+MINGW_PREFIX ?= i686-w64-mingw32-
+MINCC ?= $(MINGW_PREFIX)gcc
```
dra27, line 32: I'm not sure why all the `?=` are necessary? `make MINGW_PREFIX= ...` already overrides `MINGW_PREFIX`?

Makefile
MisterDA: We checked with Resource Hacker that when built with MS `rc`, the strings are still correct. llvm-rc cannot process the file without the quotes.

Makefile
MisterDA: Indeed, fixed!

dra27 (approved):
LGTM, thank you!

PR #155, opened 2025-02-21
Build in C99 mode in CI with cc-like compilers

MisterDA:
This is done to ensure FlexDLL can be built with old compilers or strict compiler settings.

It's not possible to use the `-Wpedantic` flag because we're casting function pointers into object pointers, or to use a strict C89 mode as we're using flexible array members (a C99 feature).

(the current little CI failure is a network problem)


dra27 (approved):
Thanks!

PR #157, opened 2025-02-28
Improve flexlink support on Unix (for cross-compiling contexts)

MisterDA:
When flexlink runs on Unix in cross-compiling contexts, it shouldn't use `cmd.exe` or batch syntax, or response files.

### DIFF
--- a/reloc.ml
+++ b/reloc.ml
@@ -204,7 +204,7 @@ let run_command cmd =
   in
   (* note: for Cygwin, using bash allow to follow symlinks to find
      gcc... *)
-  if !toolchain = `CYGWIN64 ||
+  if Sys.unix || !toolchain = `CYGWIN64 ||
      String.length cmd + String.length silencer > max_command_length
   then begin
     (* Dump the command in a text file and apply bash to it. *)
@@ -1128,7 +1128,9 @@ let build_dll link_exe output_file files exts extra_args =
         let extra_args =
           (* FlexDLL doesn't process .voltbl sections correctly, so don't allow the linker
              to process them. *)
-          let command = link ^ " | findstr EMITVOLATILEMETADATA > NUL" in
+          let command =
+            if Sys.win32 then link ^ " /nologo /? | findstr EMITVOLATILEMETADATA > NUL"
+            else link ^ " /nologo '/?' | grep -iq emitvolatilemetadata >/dev/null" in
           if Sys.command command = 0 then
             "/EMITVOLATILEMETADATA:NO " ^ extra_args
           else extra_args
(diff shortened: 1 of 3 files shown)


reloc.ml
```diff
@@ -211,11 +212,12 @@ let run_command cmdline cmd =
   in
   (* note: for Cygwin, using bash allow to follow symlinks to find
      gcc... *)
-  if !toolchain = `CYGWIN64 ||
+  if not Sys.win32 || !toolchain = `CYGWIN64 ||
```
dra27, line 215: I don't think it matters, but given that this works if flexlink is on either native Windows or Cygwin, why not instead have `if Sys.unix ||`?

reloc.ml
```diff
@@ -242,7 +244,7 @@ let quote_files cmdline lst =
   let s =
     String.concat " "
       (List.map (fun f -> if f = "" then f else Filename.quote f) lst) in
-  if String.length s >= 1024 && cmdline.may_use_response_file then Filename.quote (build_diversion lst)
+  if (Sys.win32 || Sys.cygwin) && String.length s >= 1024 && cmdline.may_use_response_file then Filename.quote (build_diversion lst)
```
dra27, line 247:
```suggestion
  if not Sys.unix && String.length s >= 1024 && cmdline.may_use_response_file then Filename.quote (build_diversion lst)
```

reloc.ml
```diff
@@ -89,6 +89,7 @@ let get_output ?(use_bash = false) ?(accept_error=false) cmd =
     else
       begin
         let (cfn, oc) = open_temp_file "longcmd" ".sh" in
+          output_string oc "exec ";
```
dra27, line 92: Is this definitely correct? The command before was going to go via `Sys.command` (i.e. "`sh -c`"-like), which this in theory breaks?

dra27 (commented):
Given that flexlink runs on both `Sys.cygwin` and `Sys.win32` already, I'd possibly change the guards to be written in terms of `Sys.unix` instead (but that's probably just paranoia).

I'm not following the need for the `exec ` bit yet (and therefore as to whether it should be guarded `if Sys.unix then` instead?)

reloc.ml
MisterDA: The command goes into a bash file that's executed as `bash longcmdxxxxxx.sh` below. As it's always a single invocation (no pipes, no command chains) and sometimes redirections, I think it's exactly the use-case for `exec`. Am I mistaken?

reloc.ml
MisterDA: Applied, thanks

dra27 (approved):
Needs a rebase, but otherwise good to go, thank you!

PR #158, opened 2025-03-27
Improve support of clang

MisterDA:
I'm trying to use clang everywhere. The MSYS2 CLANG64 environment uses UCRT, but we're missing bits to link with it right now. It is however possible to use clang with MSYS2 MINGW64 environment, or with Cygwin+mingw, or with the MSYS2 MSYS and Cygwin+cygwin1.dll environments (although in this last two cases clang packages are currently unmaintained).
It is also possible to use clang-cl instead of MSVC.

This PR makes sure that flexlink doesn't default to GCC or MSVC when clang is desired instead.

The first step is to retain the compilers used when building for searching library directories. This allows using alternative compilers, otherwise, FlexDLL would default to querying GCC, which may report different path list than clang.
The second step teaches FlexDLL to parse clang's output. `clang.exe -print-search-dirs` doesn't print an `install` field and its `libraries` path list is already in mixed-style (Windows path with forward slashes separated by semicolons). This happens when using clang in both MSYS2 MINGW64 or CLANG64 environments.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,9 +38,15 @@ CYGWIN64_PREFIX = x86_64-pc-cygwin-
 CYG64CC = $(CYGWIN64_PREFIX)gcc
 
 version.ml: Makefile flexdll.opam
-	echo "let version = \"$(VERSION)\"" > version.ml
-	echo "let mingw_prefix = \"$(MINGW_PREFIX)\"" >> version.ml
-	echo "let mingw64_prefix = \"$(MINGW64_PREFIX)\"" >> version.ml
+	echo 'let version = "$(VERSION)"' > $@
+	echo 'let mingw_prefix = "$(MINGW_PREFIX)"' >> $@
+	echo 'let mingw64_prefix = "$(MINGW64_PREFIX)"' >> $@
+	echo 'let msvc = "$(notdir $(MSVCC))"' >> $@
+	echo 'let msvc64 = "$(notdir $(MSVCC64))"' >> $@
+	echo 'let cygwin64 = "$(notdir $(CYG64CC))"' >> $@
+	echo 'let mingw = "$(notdir $(MINCC))"' >> $@
+	echo 'let mingw64 = "$(notdir $(MIN64CC))"' >> $@
+	echo 'let gnat = "gcc"' >> $@
 
 # Supported tool-chains
 
--- a/reloc.ml
+++ b/reloc.ml
@@ -24,7 +24,15 @@ let debug ?(dry_mode = {contents=false}) min_level fmt =
 let search_path = ref []
 let default_libs = ref []
 
-let gcc = ref "gcc"
+let cc = function
+  | `MSVC -> Version.msvc
+  | `MSVC64 -> Version.msvc64
+  | `CYGWIN64 -> Version.cygwin64
+  | `MINGW -> Version.mingw
+  | `MINGW64 -> Version.mingw64
+  | `GNAT | `GNAT64 -> Version.gnat
+  | `LIGHTLD -> failwith "No C compiler configured for this toolchain."
+
 let objdump = ref "objdump"
 
 let is_crt_lib = function
@@ -1173,7 +1181,7 @@ let build_dll link_exe output_file files exts extra_args =
         in
         Printf.sprintf
           "%s %s%s%s -L. %s %s -o %s %s %s %s %s"
-          !gcc
+          (cc !toolchain)
           (Option.fold ~none:"" ~some:(fun ld -> "-fuse-ld=" ^ ld ^ " ") !Cmdline.use_linker)
           (if link_exe = `EXE then "" else "-shared ")
           (if main_pgm then "" else if !noentry then "-Wl,-e0 " else if !machine = `x86 then "-Wl,-e_FlexDLLiniter@12 " else "-Wl,-eFlexDLLiniter ")
@@ -1195,7 +1203,7 @@ let build_dll link_exe output_file files exts extra_args =
         in
         Printf.sprintf
           "%s -m%s %s%s%s -L. %s %s -o %s %s %s %s %s %s"
-          !gcc
+          (cc !toolchain)
           !subsystem
           (Option.fold ~none:"" ~some:(fun ld -> "-fuse-ld=" ^ ld ^ " ") !Cmdline.use_linker)
           (if link_exe = `EXE then "" else "-shared ")
@@ -1338,7 +1346,6 @@ let remove_duplicate_paths paths =
 
 let setup_toolchain () =
   let mingw_libs pre =
-    gcc := pre ^ "gcc";
     objdump := pre ^ "objdump";
     let rec get_lib_search_dirs install libraries input =
       match input with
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -239,6 +240,15 @@ flexdll_initer_gnat.o: flexdll_initer.c
...
 	$(MIN64CC) $(GCC_FLAGS) -c -o $@ $<
 
+build_config.ml: Makefile
+	echo 'let mingw_prefix = "$(MINGW_PREFIX)"' > $@
+	echo 'let mingw64_prefix = "$(MINGW64_PREFIX)"' >> $@
+	echo 'let msvc = "$(subst ",\",$(subst \,\\,$(MSVCC)))"' >> $@
+	echo 'let msvc64 = "$(subst ",\",$(subst \,\\,$(MSVCC64)))"' >> $@
+	echo 'let cygwin64 = "$(subst ",\",$(subst \,\\,$(CYG64CC)))"' >> $@
+	echo 'let mingw = "$(subst ",\",$(subst \,\\,$(MINCC)))"' >> $@
+	echo 'let mingw64 = "$(subst ",\",$(subst \,\\,$(MIN64CC)))"' >> $@
```
dra27, line 250: I think in practice these should never end up with anything other than a compiler name, so perhaps the `subst` can be omitted?

Makefile
```diff
@@ -239,6 +240,15 @@ flexdll_initer_gnat.o: flexdll_initer.c
 flexdll_initer_mingw64.o: flexdll_initer.c
 	$(MIN64CC) $(GCC_FLAGS) -c -o $@ $<
 
+build_config.ml: Makefile
+	echo 'let mingw_prefix = "$(MINGW_PREFIX)"' > $@
+	echo 'let mingw64_prefix = "$(MINGW64_PREFIX)"' >> $@
+	echo 'let msvc = "$(subst ",\",$(subst \,\\,$(MSVCC)))"' >> $@
+	echo 'let msvc64 = "$(subst ",\",$(subst \,\\,$(MSVCC64)))"' >> $@
```
dra27, line 247: I see the intention - but, for example, it's not hardened against a path with an apostrophe - which is a legal character in a directory name - while quoting double-quotes, which are not. However - I think `$(notdir $(MSVCC))` gets us enough?

reloc.ml
```diff
@@ -24,7 +24,15 @@ let debug ?(dry_mode = {contents=false}) min_level fmt =
...
 
-let gcc = ref "gcc"
+let cc = function
+  | `MSVC -> Build_config.msvc
+  | `MSVC64 -> Build_config.msvc64
+  | `CYGWIN64 -> Build_config.cygwin64
+  | `MINGW -> Build_config.mingw
+  | `MINGW64 -> Build_config.mingw64
+  | `GNAT | `GNAT64 -> Build_config.gnat
+  | _ -> failwith "No C compiler configured for this toolchain."
```
dra27, line 34:
```suggestion
  | `LIGHTLD -> failwith "No C compiler configured for this toolchain."
```

reloc.ml
```diff
@@ -1378,13 +1381,14 @@ let setup_toolchain () =
...
-      get_lib_search_dirs "" "" (get_output "%s -print-search-dirs" !gcc)
+      get_lib_search_dirs None "" (get_output "%s -print-search-dirs" (cc !toolchain))
+      |> List.filter (( <> ) "")
       |> List.map normalize_path
       |> remove_duplicate_paths
     in
     search_path := !dirs @ lib_search_dirs;
     if !verbose >= 1 then begin
-      print_endline "lib search dirs:";
+      Printf.printf "lib search dirs (%s):" (cc !toolchain);
```
dra27, line 1391:
```suggestion
      Printf.printf "lib search dirs (%s):\n" (cc !toolchain);
```

reloc.ml
```diff
@@ -1427,16 +1430,16 @@ let setup_toolchain () =
       if not !custom_crt then
         default_libs := ["msvcrt.lib"]
   | `MINGW ->
-      mingw_libs Version.mingw_prefix
+      mingw_libs Build_config.mingw_prefix
   | `MINGW64 ->
-      mingw_libs Version.mingw64_prefix
+     mingw_libs Build_config.mingw64_prefix
```
dra27, line 1435:
```suggestion
      mingw_libs Build_config.mingw64_prefix
```
🫣

reloc.ml
```diff
@@ -1353,18 +1365,9 @@ let setup_toolchain () =
...
-                 Sys.is_directory will fail) *)
-              Filename.concat install Filename.current_dir_name
-              |> Filename.dirname
-            else
-              ""
-          in
           let separator, run_through_cygpath =
             if Sys.win32 then
-              if dir_exists_no_cygpath install then
+              if install = None (* clang *) || dir_exists_no_cygpath (Option.get install) then
```
dra27, line 1370:
```suggestion
              if Option.fold ~none:true ~some:dir_exists_no_cygpath install then
```
(I think `Option.get` then isn't needed)

Makefile
```diff
@@ -166,7 +165,9 @@ COMPAT_LEVEL := \
           $(if $(call test_ver,40500),405) \
           $(if $(call test_ver,40600),406) \
           $(if $(call test_ver,40700),407) \
-          $(if $(call test_ver,40800),408))
+          $(if $(call test_ver,40800),408) \
+          $(if $(call test_ver,40900),409) \
+          $(if $(call test_ver,41000),410))
```
dra27, line 170: I don't think this is needed in the end?

Makefile
```diff
@@ -239,6 +240,15 @@ flexdll_initer_gnat.o: flexdll_initer.c
 flexdll_initer_mingw64.o: flexdll_initer.c
 	$(MIN64CC) $(GCC_FLAGS) -c -o $@ $<
 
+build_config.ml: Makefile
```
dra27, line 243: FWIW, just update version.ml with the extra definitions (as things stand, you've not `build_config.ml` to the `clean` target but as `version.ml` already had the mingw ones there 🤷‍♀️)

dra27 (commented):
Sorry, it's a slightly nitty set of review comments - overall, I unsurprisingly like being able to add support for `clang`!

I'm mechanically suggesting to capture the basename of the command - i.e. make it snapshot the fact its `gcc`, `cl`, `clang`, etc., but don't ever allow the full path to be embedded). Given that `flexlink` is essentially built as part of the compiler these days, I think it's OK to be doing it this way, rather than adding "never to be used" mechanisms for specifying the compiler on the command line (which I think we may have already talked about offline?)

MisterDA:
I've applied your suggestions, thanks!

Makefile
```diff
@@ -38,9 +38,15 @@ CYGWIN64_PREFIX = x86_64-pc-cygwin-
...
 version.ml: Makefile flexdll.opam
-	echo "let version = \"$(VERSION)\"" > version.ml
-	echo "let mingw_prefix = \"$(MINGW_PREFIX)\"" >> version.ml
-	echo "let mingw64_prefix = \"$(MINGW64_PREFIX)\"" >> version.ml
+	echo 'let version = "$(VERSION)"' > $@
+	echo 'let mingw_prefix = "$(MINGW_PREFIX)"' >> $@
+	echo 'let mingw64_prefix = "$(MINGW64_PREFIX)"' >> $@
+	echo 'let msvc = "$(notdir $(MSVCC))"' >> $@
+	echo 'let msvc64 = "$(notdir $(MSVCC64))"' >> $@
+	echo 'let cygwin64 = "$(notdir $(CYG64CC)"' >> $@
```
dra27, line 46:
```suggestion
	echo 'let cygwin64 = "$(notdir $(CYG64CC))"' >> $@
```

MisterDA: Fixed! thanks

dra27 (approved):
LGTM, thank you!

PR #160, opened 2025-10-17
Constify parameters

MisterDA:
An incredibly useful patch ;)
The binaries are byte-for-byte identical between 0.44 and this patch with GCC 15.2.


nojb (approved):
LGTM

nojb:
Thanks!

PR #161, opened 2025-10-17
Remove unused `intnat`/`uintnat` definitions

MisterDA:
Their names are confusing with respect to the OCaml runtime definition of `intnat` and `uintnat`, which are wide enough to store a pointer. On 64-bits, FlexDLL's `uintnat` remains a `long`, thus 32 bits wide. In all other places the truncation of the pointer is done by casting to a `UINT32*`, do that instead.

### DIFF
--- a/flexdll.c
+++ b/flexdll.c
@@ -27,9 +27,6 @@
 #endif
 #endif /* MSVC */
 
-typedef long intnat;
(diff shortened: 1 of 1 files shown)


nojb (approved):
LGTM

PR #164, opened 2026-01-08
Support -mwindows and -mconsole for greater compatibility with pkg-config

nojb:
Attempt in the direction of https://github.com/ocaml/flexdll/issues/163#issuecomment-3721857131.

Fixes #163

PR #165, opened 2026-01-15
AppVeyor updates

dra27:
Four easy changes:
- Add testing for 5.4 and, for variety, test msvc64 by switch to the AppVeyor VS 2022 image
- Switch the trunk build to the VS 2022 image to fix [#164 (comment)](https://github.com/ocaml/flexdll/pull/164#issuecomment-3728911042)
- Switch 5.3 to test on MSVC as well (as the older 4.x images do)
- Add labels to the matrix items

Then the rabbit-hole:
- Deal with `safe.directory` 🥱
- Update the bootstraps to download FlexDLL 0.44, or [ocaml/ocaml#14034](https://github.com/ocaml/ocaml/pull/14034) will bite at some point!
- Actually activate the appropriate version of Visual Studio on the VS 2022 images!
- Update the i686 mingw demo to cope with mingw-w64 runtime 8.0.0+ (cf. [ocaml/ocaml#10046](https://github.com/ocaml/ocaml/pull/10046) - just hadn't noticed because the default AppVeyor image we use is intentionally old)
- Display a few more Cygwin package versions to confirm the previous diagnosis

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -252,8 +252,10 @@ demo_msvc: flexlink.exe flexdll_msvc.obj flexdll_initer_msvc.obj
 demo_cygwin64: flexlink.exe flexdll_cygwin64.o flexdll_initer_cygwin64.o
 	$(MAKE) -C test clean demo CHAIN=cygwin64 CC="$(CYG64CC)" CFLAGS="$(GCC_FLAGS)" O=o RUN="PATH=\"/cygdrive/c/cygwin64/bin:$(PATH)\""
 
+# cf. ocaml/ocaml#10046
+MINGW_DEMO_CFLAGS=-static-libgcc
 demo_mingw: flexlink.exe flexdll_mingw.o flexdll_initer_mingw.o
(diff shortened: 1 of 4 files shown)


nojb:
Wonderful. Thanks @dra27!

nojb:
> * Switch the trunk build to the VS 2022 image to fix [#164 (comment)](https://github.com/ocaml/flexdll/pull/164#issuecomment-3728911042)

Can you say a few words about this? I thought the problem was with the `mingw64` headers somehow?

MisterDA (approved):

dra27:
It is indeed the headers - the Visual Studio 2015 image we're using by default isn't being updated by AppVeyor, so it's got a very old Cygwin installation... the alternative would be to update the Cygwin installation in it, but that's not trivial to do (cf. https://github.com/dra27/ocaml/commit/64a2137ada6ee5ec5d3aa23476cd969fec2519d5) and of course we really ought to be testing the newer image as well!

PR #166, opened 2026-05-03
Fix executable permission on scripts and non-scripts

MisterDA:
Very minor patch, I just noticed that two files had the executable permission set which could be unset, and one script did not have it.


nojb (approved):
LGTM

PR #169, opened 2026-06-15
Drop the `.llvm_addrsig` section from rewritten objects

MisterDA:
The address-significance table emitted by Clang holds symbol table indices, which become stale when flexlink rewrites an object's symbol table to add the relocation table. lld-link rejects objects with an invalid `.llvm_addrsig` section. The section is an optional hint for the linker's identical-code-folding pass and can safely be dropped.

It is also possible to tell Clang not to emit this section with `-fno-addrsig`.

Proper handling of this section could come later, for now it's easy to silently ignore it.

See also:
- https://llvm.org/docs/Extensions.html#sht-llvm-addrsig-section-address-significance-table
- https://clang.llvm.org/docs/UsersManual.html#cmdoption-faddrsig
- https://static.googleusercontent.com/media/research.google.com/en//pubs/archive/36912.pdf


CHANGES
```diff
@@ -2,6 +2,8 @@ Next version
 - GPR#164: Recognize `-mwindows` and `-mconsole` for greater compatibility (with
   `pkg-config` in particular).
   (Nicolás Ojeda Bär, review by David Allsopp)
+- GRP#168: Drop the .llvm_addrsig section from rewritten objects.
```
MisterDA:
```suggestion
- GRP#169: Drop the .llvm_addrsig section from rewritten objects.
```

reloc.ml
```diff
@@ -926,11 +926,29 @@ let build_dll link_exe output_file files exts extra_args =
...
+           | `Num _ -> true)
+        obj.symbols
+  in
+
   let record_obj obj =
     if !builtin_linker then ""
     else begin
       let fn = temp_file "dyndll" (ext_obj ()) in
       let oc = open_out_bin fn in
+      drop_addrsig obj;
```
dra27: Is this the "right" place to do this? I'm not certain, looking at it, but doesn't this get called for both rewritten objects _and_ the descriptor, where `close_obj` is only called for the former? If that's right, wouldn't it be technically better to do this stripping there, where we're definitely copying an existing object, and not for the descriptor object, which is entirely generated?

MisterDA: Indeed! I've moved it, thanks.

dra27 (approved):
LGTM, thanks!

